PR #12, opened 2016-07-11
Fix project for use within virtualenv

mbrgm:
- Use /usr/bin/env to determine path to python3 instead of a hard-coded
  path, so that virtualenv stubs work properly.
- Change package name to 'boxes' ('boxes.py' did not work within
  virtualenv).


florianfesti:
This patch should change all the other #! lines, too. At least the one in boxesserver.

mbrgm:
Fixed the remaining shebangs.

PR #13, opened 2016-07-11
Use cairocffi instead of cairo

mbrgm:
cairocffi is available from pip, which makes setup much easier.


florianfesti:
Merged with some try block to keep working with the old pycairo lib.

PR #14, opened 2016-07-16
Adds 3 more box projects to generators

bobm123:
First of all thanks so much for writing this library. I was about to attempt something like it myself, but fortunately Googled and found yours first. As an "exercise for the student" I took your template and tried to generate a box of my own. I knew I needed some of my own parameters and special edges for the projects I had in mind and that seemed like a good way to try out your library without getting too crazy.

Next up was a caddy.py, a simple 6-pack carrier. I also plan to make one of these sized to carry 4 2-liter bottles to the party. That was fun and I think it might be useful somebody. I intended these to be used with trayinsert.py but an integrated handle might be nice here too.

Finally there is the thing I call portmanteau.py from the kind of luggage of a century ago. In my case I needed such a case for my daughter's microscope. The cardboard box that it came in is falling apart and I hope this will be a more suitable replacement. The idea is to make two sets of these parts then join them on the long edges with a piano hinge.

I will be back at my local makerspace in a couple days and will try to make some test cuts. I expect there there may be adjustments. I will you know how it turns out. This is my first attempt at a pull request to an open source project so I would like to hear any comments or suggestions you might have.

Thanks,
Bob M


florianfesti:
Thank you very much for submitting your code. I will treat this more of an request for comment than an actual pull request as some aspects will need some more work before they can be merged in.

Obviously the upper two walls of the portmanteau still need finger joints. Try Boxes.fingerHolesAt() which is an instance of edges.FingerHoles. I would also suggest to include both sides in the generator. Also try placing the side walls more nicely with their 45° part overlapping to save space.

You may want to consider changing the edge types for the bottom of the caddy to "f" for stability reasons. You can then use "h" for the bottom of the side walls or even "s" to get feet that will keep moisture from your caddy and make it stand more stable on uneven ground.

To simplify the example you can remove the bedbolts=None parameters.

florianfesti:
In general should separate features - like separate generators - be submitted as separate pull requests. This as the benefit of making review easier and also allows getting some things upstream while others still need work or discussion. I won't make this a requirement here but just as a general note.
More important IMHO is submitting clean patches. Fixes should be squashed into the patches that introduce the code. All patches that go upstream should not have known defects. I typically run git rebase -i many times on a feature branch before adding it upstream. I am also not a big fan of merges (but different people have different opinions) and basically do not use them at all. Instead I keep my local branches rebased on top of master which gives a much cleaner history. Patches also should only contain changes for one feature or topic. Mixed up patches can be split up by selecting them for editing in git rebase -i, then git reset HEAD^ and then recommitting the part separately - if necessary using git add -i. But this requires some proficiency with git. So if you need help with cleaning up the patches just ask me.

As the generators are pretty small and straight forward they should probably be submitted as three patches - one for each one. More complicated features or ones that also add to the library may be split up into multiple patches each containing a sub feature or representing a significant stage in the development.

bobm123:
Thanks for the feedback. It was a very busy week at the makerspace, so I only had a chance to make a couple of test cuts. Even so, I did find some problems with the portmanteau project. Once I get those changes retested and your suggestions incorporated, I'll resubmit separate pull requests for the new file.

I really like your idea of using feet at the bottom of the caddy. Since one of the planned uses is to make totes for carrying bottles of sugar water to my beehives, these will be very helpful.

florianfesti:
See bottom_edge parameter of box2 as an example on how you could make the bottom edge of the caddy configurable.

florianfesti:
Ok, this PR has been open for much too long. At first glance the three new generators look good. They may benefit from a bit of polishing but this really should not delay merging them any further.
One way would be to just merge the branch. Another might be just adding each generator in a single commit. I'll do what ever you prefer.

bobm123:
Sorry I've let this one hang around so long. I'll try to clean it up a bit and make separate submissions as you suggested a while back. If all goes well I should be able to get to this in the next day or two.

Thanks!
Bob M

florianfesti:
OK, closing this now. Please open one PR per generator.

florianfesti:
I had a bit time for looking at your branch while being on the train. If there's no objection I'd like to add some of the generators as new files without any history. If this is fine with you I'd add you in the copyright section on top and set you as author of the patch adding the file.
For the single generators:
Caddy: I would probably rather add the RoundedTriangleEdge to the TypeTray. As this basically just affects two edges that should be pretty easy. I already made the bottom_edge configurable to add stability.
Magazinefile_Ogee: Either add or merge as option to the regular MagazinFile (haven't looked too closely)
Portmanteau: Add. May be add optional hinges (as in HingeBox). Render both side - may be with different widths. I will rearrange the parts a bit to better fit on the sheet.
TwoPiece: I basically reworked the generator. Fixed the todos. I'd add it as is but put a patch on top basically replacing most lines. But this is more about the idea than the implementation. As is is basically just two boxes used in a clever way. 
Haven't looked at example.py and drawer.py yet. They are lower on my priority list.

florianfesti:
OK, I finally managed to get https://festi.info/boxes.py/TwoPiece up and running. I rewrote most of the code but this is more about the idea. The result is very nice. Thanks for submitting the idea. I mentioned you in the commit message. I hope that OK.

florianfesti:
OK. I finally added the RoundedTriangleEdge from the caddy generator. It is now supported for both the UniversalBox and the TypeTray (which are the only generators using the top_edge param). The later should basically give the same result as the caddy.
See
31fd0024eedcc9ed02f148eb3093acc43e909c0c
2747d9719a21a0277fab446327f4b4a4abeece61

Issue #48, opened 2017-11-18
Description for generators


DanielaFesti:
It would be cool if the generators could have descriptions (longer than just the one liner in the __doc__ string).

The descriptions could contain information for assembling the generated box, or lists of materials when the thing you build contains more than just the generated walls.

For nicer formatting, a markup language would be good like Markdown (as used by github). Maybe this package would work? -- https://pypi.python.org/pypi/Markups

If projects get more complex, they often get their own homepage somewhere else, e.g., on hackaday.com. Hence, they also should get an (optional) URL that links to more details.

florianfesti:
New code uses the markdown library.

PR #49, opened 2017-11-18
Data structure for description and url

DanielaFesti:
This is how description and url could be added to box generators (see Issue #48).

The added variables are not rendered yet.


florianfesti:
Merged.

PR #55, opened 2018-01-16
Fix tmp file access in windows

deepinthewoods:
Was throwing this error:
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\n\\AppData\\Local\\Temp\\tmpw8emq0xq'

see https://stackoverflow.com/questions/34716996/cant-remove-a-file-which-created-by-tempfile-mkstemp-on-windows


florianfesti:
Thank you very much for your effort of getting Boxes.py running on Windows!

PR #56, opened 2018-01-18
 enable bedbolts for Straight 'e' and OutSet edges 'E' 

stegro:


florianfesti:
Thanks for the patches. I fixed the tabs in an additional patch afterwards.

stegro:
Great! In fact I ignored them since I don't see what they mean and there is no comment on them. Can you say a few words, about what 'tabs' are?

florianfesti:
They are gaps in the outline of parts that create bridges that keep the part from falling out. This makes it easier to ship and store the kits as it stays as a sheet. It can be controlled by the tabs param in the Default Settings. Values of 0.2 to 0.25 work well with our laser cutter.

But yes, some comments would be good.

PR #67, opened 2018-06-14
boxesserver: Implemented toggling of list item icon

stefan2904:
Triangles before list items now toggle according to open/closed state of item.

20180614_selection_02

(Also removed some random whitespace, sorry.)


florianfesti:
Thanks for the patch set!

PR #68, opened 2018-06-22
reduce size of images to speed up website

widdakay:
This seems to be the source of the images which are loaded on the website. On many slower and cell connections, these images take a long time to load.


Rotzbua:
I agree with your intention to speed up the page but I disagree with your solution.

Your solution introduces a second picture, which has to be generated manually every time a new or updated picture is pushed to the repository. This increases the work and maybe cause inconsistency.

Better would be a script which reduces the size of the pictures. The script can be used in a CI, deploy script or manually at server deployment.

florianfesti:
Sorry for this taking so long. I am still undecided on how big the pictures should actually be.

Also Rotzua is kinda right that we should check in a script that can generate the thumbnails.

PR #80, opened 2018-08-28
Change the generate script

pinkfish:
Change the generate script to run from the command line, fixing the changes in the parameters.


florianfesti:
Wow! I haven't looked into that script for years. Thanks for cleaning up the mess.

PR #92, opened 2018-10-05
Add a box for storage of playingcards

MrShark:
The box have place for two stacks of cards with a divider between.
It have a slide in lid.

Not tested in a cutter yet, but seems correct.

### DIFF
--- a/boxes/generators/cardbox.py
+++ b/boxes/generators/cardbox.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+# Copyright (C) 2013-2014 Florian Festi
+# Copyright (C) 2018 jens persson <jens@persson.cx>
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from boxes import edges, Boxes
+
+
+class InsetEdgeSettings(edges.Settings):
+    """Settings for InsetEdge"""
+    absolute_params = {
+        "thickness": 0,
+    }
+
+
+class InsetEdge(edges.BaseEdge):
+    """An edge with space to slide in a lid"""
+    def __call__(self, length, **kw):
+        t = self.settings.thickness
+        self.corner(90)
+        self.edge(t, tabs=2)
+        self.corner(-90)
+        self.edge(length, tabs=2)
+        self.corner(-90)
+        self.edge(t, tabs=2)
+        self.corner(90)
+
+
+class FingerHoleEdgeSettings(edges.Settings):
+    """Settings for FingerHoleEdge"""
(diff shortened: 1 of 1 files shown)


florianfesti:
Looks very nice!

Minor nit pick: The lidwidth should probably not be a parameter but be hard coded to one thickness. That way the lips won't protrude into the airspace above the cards.

One would probably have to actually cut it out to see if the finger holes are wide enough (depending on material used) to really pick up the cards conveniently. Such a small box can easily be cut out of 3mm ply which leave very little room for clumsy fingers like mine.

boxes/generators/cardbox.py
```diff
@@ -0,0 +1,153 @@
...
+        self.ctx.restore()
+        self.rectangularWall(x, h, "EEEE", move="up only")
+        self.ctx.save()
+
+        # Outer sides
+        self.rectangularWall(h, y, "fFfF", move="right")
+        self.rectangularWall(h, y, "fFfF", move="right")
+
+        # Inner sides
+        self.rectangularWall(h, y, "eAee", move="right")
```
florianfesti: These need to be h-t in height - as is divider already is.

boxes/generators/cardbox.py
```diff
@@ -0,0 +1,153 @@
...
+        self.addPart(p)
+
+        s = FingerHoleEdgeSettings(thickness=t, wallheight=h)
+        p = FingerHoleEdge(self, s)
+        p.char = "A"
+        self.addPart(p)
+
+        self.ctx.save()
+
+        # Lid
```
florianfesti: The lid is a bit too wide. It is a press fit as all parts are by default. The width need to be reduced to allow some play and easy movement.0.1*t or even 0.2*t may be a good value.
At the same time the lid is not deep enough. It should overlap the over wall. So it needs to be y+t deep.
I would probably also add a finger hole to make sliding the lid out easier.

boxes/generators/cardbox.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Outer sides
+        self.rectangularWall(h, y, "fFfF", move="right")
+        self.rectangularWall(h, y, "fFfF", move="right")
+
+        # Inner sides
+        self.rectangularWall(h, y, "eAee", move="right")
+        self.rectangularWall(h, y, "eAee", move="right")
+
+        # Lips
```
florianfesti: The lipwidth parameter should go away. Just use t for the width of the lips. The 10mm default basically makes it impossible to use the finger holes at the sides.

boxes/generators/cardbox.py
florianfesti: Having the back (or all) corners of the lid rounded with a small radius would make inserting the lid much easier.

florianfesti:
I should be able to re-cut the changed parts this evening and confirm that this fixes the issues.

florianfesti:
With the fixes this results in a working box.
If you want I can squash the two patches together and commit the result upstream. Otoh there may be some more tweaks you want to add before getting the box upstream.

There are two more improvements I came up with:
1) Have a param for the number of decks. This box would also work well with only one and also up to may be 4 compartments.
2) Add a spring loaded indents similar to the slide on edge to keep the lid in place. They could be part of the inner sides.
But they can both wait until the basic generator is upstream.

MrShark:
Thanks, I don't think I will have time doing tweaks the coming week, so if you like to squash and commit do so (I did the squash locally, but github is not gerrit so I could not get it pushed without reading the manual :-)

florianfesti:
Ok, squashed and pushed. Thanks for the work!

florianfesti:
Added a num param in 5b93c1228d718e7a0e94ff3c3fb44095d6df7ac2

PR #94, opened 2018-10-21
Add DisplayCase generator

abulimov:
Very simple generator based on OpenBox.
I use it to cut display cases for my scale models.


florianfesti:
Merged. Thanks for the patch.

I wonder if it wasn't more practical to make to glass top removable. May be by not having finger joints on the bottom but have two rectangles at the bottom glued into each other to create a step the top case can sit around.

PR #96, opened 2018-10-27
Add a generator for paintstands

MrShark:
A simple stackable paintstand, very much inspired by:
https://www.youtube.com/watch?v=6HJarJg1npk

### DIFF
--- a/boxes/generators/paintbox.py
+++ b/boxes/generators/paintbox.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+# Copyright (C) 2013-2016 Florian Festi
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


florianfesti:
Very nice! Also a very nice use of the already existing features! From my POV this is ready to be merged. I still have a few idea of things you may want to improve beforehand. If not just drop me a note and I'll hit the merge button.

Add yourself to the copyright notice

Adding a vertical wall connecting the two layers along the lengths would make the stand much stronger. Don't know if this is really needed. But after hearing all the talk about squaring stuff up in the video... Even having it for only a short length would keep the stand from racking side by side. Needs some extra code to not collide with the holes for the bottles, though.

You can add an option to use .hexHolesRectangle() to squeeze an additional row onto narrow paintstands.

MrShark:
Thanks, I'm out of town for a week and I can't do patches, so just now so go ahead and decide.
I have cut a few stands and I don't think a wall is needed, but have an idea for a matching tray for keeping brushes and other things with sides. Perhaps adding sides and making the "hole part" optional could be something.
HexHolesRectangle sounds good, I have done it manually on some of my stands. Will look into it if you don't beat me to it 😁

florianfesti:
Well, then let's get this in right now and add all the other features afterwards.

florianfesti:
Rearranged the pieces a bit to make them better fit together.
Also added the hex holes.

PR #105, opened 2018-12-12
Added ponoko-specific changes to color and line width.

wayn3w:
This feature adds support for creating boxes destined for Ponoko.  The user's workflow is to specify
 scripts/boxes <generator>  --format=svg_Ponoko
The user will then import the resulting svg into one of the three Ponoko-supplies templates.

In order to support this feature, I used a python generator for saving and restoring the cairo context around color changes so there won't be a need to remember the previous color.  I kept other refactoring to a minimum.


florianfesti:
That's clearly going in the right direction.

I would prefer if self.ctx would be encapsulated with both the saved() context manager and the new color handling.
While I have not yet decided on replacing cairo as a graphics library I would like to see cairo specific being removed from the API. Those two are basically the only place left where the ctx object needs to be used by generators. Everything else is hidden in some Boxes methods for drawing stuff.

This should not be too difficult. Just make saved a method of the Boxes class and add a method for setting the color.
I would probably also have this as three separate patches: new saved method, color handling and new format. But I am fine to split that up myself if you don't feel comfortable with the git magic needed to do that.

I would probably move all the generator over to the new saved method. But this is out of scope here and can be done later on.

wayn3w:
I would be happy to take your direction on rewriting my changes.  I can
implement it as three separate pull requests to be considered serially
because I don't have as much familiarity with git to do it any other way.
If you see a different way of doing this, please point me to more info; I
appreciate the instruction.

Thanks!  I'm looking forward to this.

wayn3w:
I have created a new pull request.  I hope it is to your liking; as for the
other things you wanted to see done, I'm happy to do them after this is
approved.  I was concerned about regressions, since there's a lot of code
without test cases.  If this is accepted, I'd like to propose another
feature.

Thanks for your lending your experience and giving me the opportunity to
learn.

Wayne

florianfesti:
Ok, this PR is superseded by #112 

Closing here.

PR #112, opened 2019-01-22
Refactored context management; Adding Ponoko -specific line properties.

wayn3w:
This is the the same as the other Ponoko-specific pull request, but split into separate commits.

Not done:  
- Changing the generators to use saved_context().  Happy to do this as a separate code change.
- Adding restoring the original point (get_current_point, move_to) to saved_context().  This had the potential to change a lot of code, including removing Boxes.restore().  Happy to do this as a separate code change.


florianfesti:
Thanks for the updated patch. I am currently traveling so it may take a few days longer to take a look at it.

florianfesti:
Ok, this looks really good. I split the patches even more (separating the color from the save_context part and moving the GPL header into a separate patch). I'll do some more testing tomorrow and then push the whole set.

Minor nit picks so far: The changes to universalbox change the color from red (1.0, 0, 0) to blue. 
Looks like formats needlessly imports Colors
But I'll just fix that myself.

wayn3w:
Great! Would you like me to refactor the generators to use the save_context
once you merge?

Wayne

wayn3w:
If there's something in particular you'd like me to do re: this update, let
me know; otherwise I'm going to work with another concept I've been
thinking about.

Wayne

florianfesti:
Ok, I fixed the issues above and simplified the handling of the svg_Ponoko format a bit. I still kept you as sole author. I hope that's ok.

wayn3w:
I have no objections to you being co-writer, or giving me sole credit.  I'm
just happy to contribute.  Thank you.

Wayne

florianfesti:
Ok, I pushed my changes to switch over to saved_context(). Still a few generators left that need more than just a few lines indented on level deeper.

PR #116, opened 2019-02-25
add moving; contextmanager for self.move (before and after)

foosinn:
Since self.move() is called before and after a block a context manager can be used.


florianfesti:
Yeah, I looked into using a context manager for this already. It would be really nice if it would work. But so far it looks like there is not sane way to skip the block inside the with statement.

I am closing this PR. If someone finds a sane way to do this, please open a new PR.
AngledBox and FlexBox2 may be testing grounds for the next attempt.

PR #119, opened 2019-02-26
Add installation guide for Windows (fixes #83)

### LINKED ISSUE #83: Installation Docs for Windows
Setting up Boxes.py on Windows is not entirely trivial. Main issue is getting the cairo library working. Things are even more complicated if things are expected to work with Inkscape. There are various bits of information distributed in different tickets but no clear and tested instructions.

If some actually using Windows and got Boxes.py running (may be even with Inkscape and every bit helps) could provide instructions that be great.

muellermartin:
I also have some screenshots which might help Windows users even more. If you like having them included, simply drop a line.


florianfesti:
Screen shots would be really nice. I wonder if it might be better to move the Windows instructions to a separate page then. But I can also do that after the screen shots were added.

PR #125, opened 2019-03-06
inkscape install documentation simplified

wuschLOR:


florianfesti:
Thanks for the patch. Pushed after rebasing and squashing.

PR #132, opened 2019-04-06
DiscRack

chrysn:
I'm implementing a shelf-like "box" that can be used to store disc-shaped objects vertically in a rack.

The current branch is missing some checks (e.g. whether the rear grid and the lower grid actually connect inside the side walls, whether any grid protrudes on the other side, whether slits exceed the grid area or whether the discs can still be pulled out), and I have not cut & assembled an object from it yet.

Once those checks are added, there is a picture to go with it, and I've split the clean-up commits from the actual implementation, I'd like to suggest DiscRack for inclusion in the collection.

### DIFF
--- a/boxes/generators/discrack.py
+++ b/boxes/generators/discrack.py
@@ -80,27 +80,27 @@ class DiscRack(Boxes):
         super().__init__()
         self.buildArgParser("sx")
         self.argparser.add_argument(
-            "--disc-diameter", action="store", type=float, default=150.0,
+            "--disc_diameter", action="store", type=float, default=150.0,
             help="Disc diameter in mm")
         self.argparser.add_argument(
-            "--disc-thickness", action="store", type=float, default=5.0,
+            "--disc_thickness", action="store", type=float, default=5.0,
             help="Thickness of the discs in mm")
 
         self.argparser.add_argument(
-            "--lower-factor", action="store", type=float, default=0.75,
+            "--lower_factor", action="store", type=float, default=0.75,
             help="Position of the lower rack grids along the radius")
         self.argparser.add_argument(
-            "--rear-factor", action="store", type=float, default=0.75,
+            "--rear_factor", action="store", type=float, default=0.75,
             help="Position of the rear rack grids along the radius")
 
         self.argparser.add_argument(
-            "--disc-outset", action="store", type=float, default=3.0,
+            "--disc_outset", action="store", type=float, default=3.0,
             help="Additional space kept between the disks and the outbox of the rack")
         self.argparser.add_argument(
-            "--lower-outset", action="store", type=float, default=0.0,
+            "--lower_outset", action="store", type=float, default=0.0,
             help="Space in front of the disk slits (0: automatic)")
         self.argparser.add_argument(
-            "--rear-outset", action="store", type=float, default=0.0,
+            "--rear_outset", action="store", type=float, default=0.0,
             help="Space above the disk slits (0: automatic)")
 
         self.argparser.add_argument(
--- a/boxes/generators/discrack.py
+++ b/boxes/generators/discrack.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+# Copyright (C) 2019 chrysn <chrysn@fsfe.org>
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


florianfesti:
This looks promising! The NotImplementedError thing can be merge upstream right away as soon as you split it out.
I will probably fix the AllEdges thing in a different way. The Edege objects already have the .spacing() method. So this can be done in a generic way for all edge types. I wanted to add some short edges around the corners right and left to make it easier to understand which way the edges are pointing.
Wrt the actual generator:
You need to replace the "-" in the param names by "_" to make things work. I am kinda surprised that this did even render on your side.
I am also wondering if you can't get away with less params. May be some things can be the same bottom and at the back. Or you can use good defaults. Yes, there is a thin line between not offering enough configurability and overloading the users with unnecessary choices.
Just out of curiosity: What type of disks are we talking about here. I am a bit puzzled about the huge default gaps between the slots.

chrysn:
On Sun, Apr 07, 2019 at 03:31:57AM -0700, Florian Festi wrote:
> The NotImplementedError thing can be merge upstream right away as soon as you split it out.

I'll pick that into a dedicated PR.

> I will probably fix the AllEdges thing in a different way. The Edege
> objects already have the .spacing() method. So this can be done in a
> generic way for all edge types. I wanted to add some short edges
> around the corners right and left to make it easier to understand
> which way the edges are pointing.

That sounds much better; for me that was a good starting point to get
familiar with how painting happens in boxes in general.


> You need to replace the "-" in the param names by "_" to make things
> work. I am kinda surprised that this did even render on your side.

I only use boxes on the command line so far -- and there, argparse
conveniently allows to have dashes in the arguments (as it is customary
on UNIX command lines), and exposes them with underscores in the parsed
data (as it is usable with Python).

I'll test with the web and the inkscape frontend to see if anything
breaks there, but can use underscores even if things work if that's the
common style in boxes.py.

> I am also wondering if you can't get away with less params. May be
> some things can be the same bottom and at the back. Or you can use
> good defaults. Yes, there is a thin line between not offering enough
> configurability and overloading the users with unnecessary choices.

It would make sense to make the {lower,rear}-offset parameters automatic
in the sense that they always extend to the edge of the box. Can
parameters have a None sentinel value without breaking non-commandline
user interfaces, and if so be grouped as "extended customization
options"?

> Just out of curiosity: What type of disks are we talking about here. I
> am a bit puzzled about the huge default gaps between the slots.

The huge default gaps come from the default value of sx, but actually
fit very well with my use case of kitchen appliance slicers similar to
those at [1].

[1]: https://duckduckgo.com/?q=kenwood+slicer&t=ffab&iax=images&ia=images&iai=https%3A%2F%2Fi.ytimg.com%2Fvi%2FiDgmL7je_gc%2Fmaxresdefault.jpg

chrysn:
Now with some clean-up (still untested in practice, so not removing the WIP flag):
* I've removed two configurable parameters and pull the inner parts all to the edge of the sides instead.
* There are checks now for all the things that can go wrong if the lengths don't add up.
* There is now a sane default value for sx (the global default is not suitable as you pointed out; turns out I did need to override them all the time, and just erred in my previous comment)
* Works now from web frontend (dashes replaced with underscores)
  * As multiple errors can occur with this box, I've added multi-line capability to the web server.
* Python 2 compatibility restored

It Would Be Nice™ to have the errors show up as warnings and to have the disabled parameters in some kind of "extended configuration" expander, but that's something that can still be added later globally.

florianfesti:
Any news on this? Is there anything I can do to help getting this ready for being merged upstream?

chrysn:
On Sun, May 19, 2019 at 07:05:14AM -0700, Florian Festi wrote:
> Any news on this? Is there anything I can do to help getting this
> ready for being merged upstream?

I haven't gotten around to testing the output yet (just bought the wood
for it yesterday), and am not really comfortable with asking for a merge
review when I've not even assembled one yet.

Short of that check, the code is ready from my POV, and I'll update this
issue once cut, assembled and (possibly) photographed.

florianfesti:
Cool, thanks for the update!

chrysn:
The rack could be assembled well, I'd consider this ready. I'll file a picture later.

(I've got some new ideas on what could be improved -- the sides could be diagonally cut off or be rounded to match the discs, but that's for another time.)

florianfesti:
OK, I squashed most of the patches and pushed them upstream. Thanks for the code. Submitting a nice picture would be appreciated.

PR #142, opened 2019-05-13
Add a StorageTray generator. This is a box (shelf) for storing TypeTray boxes.

Shikoruma:
It is based on TypeTray generator. The sh parameter replace the sx parameter.
It is not a subclasse of TypeTray in order to have the top and bottom at
the right place.


florianfesti:
Ok, this looks good. I am not too keep on moving the long description from --sx to --sy but I'll probably just merge anyway and figure out something better later on. The long format description should probably move to the docs with there only a pointer remaining for all section params. But that's out of scope here.

I think the use of top_edge makes the UI much more complicated than it should be. One could argue that the top and bottom edges should not be selectable but just be straight "e" Edges. The alternative would be a custom param that only allows one of "efS" or something very similar mirroring the bottom edge. This way at least the edge setting for the more exotic stuff can be removed.
There also is no real point in having a lid on top as everything is accessed from the front.

What about renaming the generator to StorageShelve and move it to the Shelves category as this make more clear that it is to be used upright?

Shikoruma:
I totally agree for the name. 

For the edge I think it's important to keep all of them. We can't see the use but some day someone will need it.

I make the change.

Shikoruma:
Changes done, if you think something else, tell me.

florianfesti:
Ok, squashed and merged.
One thing that might need improvement is the handling of the `outside`parameter for the height. Depending on the edge types used the height needs different corrections. That's the reason the `outside` parameter is currently not supported for the universal box.
When arbitrary edge types have only be added to the TypeTray this should have been resolved but it has not. So this generator inherits this problem. But that's not really a good reason to keep this out of the upstream repo . I might fix this in all three generators at once at some point.

PR #153, opened 2019-07-20
forward holesSettings to hexHolesRectangle from within rectangularWall

MAKOMO:
Shouldn't the holeSettings given to rectangularWall be forwarded to the inner hexHolesRectangle call?

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1703,7 +1703,7 @@ class Boxes:
         if holesMargin is not None:
             self.moveTo(holesMargin,
                         holesMargin + edges[0].startwidth())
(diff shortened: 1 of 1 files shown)


florianfesti:
Overall both the holesSettings as a tuple and the whole idea of those walls supporting hexholes on their own (instead of the caller using callbacks) are on the list of things to clean up. Nevertheless this is not a reason not to fix this now. Just don't grow too attached to the mechanism and its implementation details.

florianfesti:
Anyway, thanks for the patch!

MAKOMO:
Dear Florian, thanks for your lib and the hint on using the callbacks. Works fine that way too.

PR #155, opened 2019-07-26
Lid fixes

matthijskooijman:
This PR fixes a number of problems in the way burn and play were handled
for lids. These bugs would cause misalignment between the two pin slots
that anchor the lid and between the lid and the finger holes where it
should slide in.

To illustrate these problems, I'm using the universalbox generator, with
the following patch applied. This patch skips one wall, which causes the
back wall and one side wall to end up adjacent, which makes it trivial
to check the alignment of the pin slots in the side and the finger holes
in the back.

```patch
--- a/boxes/generators/universalbox.py
+++ b/boxes/generators/universalbox.py
@@ -84,8 +84,8 @@ class UniversalBox(_TopEdge, _ChestLid):
 
         self.rectangularWall(x, h, [b, "F", t3, "F"],
                              bedBolts=[d2], move="right only")
-        self.rectangularWall(y, h, [b, "f", t2, "f"],
-                             bedBolts=[d3], move="up")
+        #self.rectangularWall(y, h, [b, "f", t2, "f"],
+        #                     bedBolts=[d3], move="up")
         self.rectangularWall(y, h, [b, "f", t4, "f"],
                              bedBolts=[d3], move="up")
 
```

With this patch applied, you can reproduce the problem with the finger and
spring hole positions due to burn with:

    $ boxes Universalbox --x 30 --y 30 --top_edge=L --bottom_edge=e --burn=.2 --Lid_play=0                                                                    

image

Note the (blue) finger hole on the right which shifted, as well as the (blue)
spring hole on the left. Also note that the position of the two pin slots on
the left (in black) is correct, 2x3mm from the top.

To reproduce the problem with space/play, run:

    $ boxes Universalbox --x 30 --y 30 --top_edge=L --bottom_edge=e --burn=0 --Lid_play=0.1                                                                   

image

Here, the guides are placed in the canonical position of the lid (e.g. without
any play applied). You can see that the play is added below for the left slot
and above for the right slot. Note that the play is evenly distributed for the
(blue) hole (whose position is correct because the burn is set to 0).

I also added a commit (the first) that slightly improves some comments,
and added a commit that lets edge_width be consistently be interpreted
(the back wall / finger holes edge already did, now the side walls also
do, which allows added a bit of extra space above the lid, which can be
useful when using thing material).

With this PR applied, both burn and play result in a symmetrical aligned result:

    $ boxes Universalbox --x 30 --y 30 --top_edge=L --bottom_edge=e --burn=0.2 --Lid_play=0.1

image

Note that the finger holes are slightly larger, IIRC the holes apply the play on both sides in whole, while the pin slots divide the play over both sides.

### DIFF
--- a/boxes/edges.py
+++ b/boxes/edges.py
@@ -1430,6 +1430,8 @@ class LidSettings(FingerJointSettings):
 
     """Settings for Slide-on Lids
 
+Note that edge_width below also determines how much the sides extend above the lid.
+
 Inherited:
 
     """
@@ -1532,6 +1534,7 @@ class LidSideRight(BaseEdge):
         t = self.boxes.thickness
         s = self.settings.play
         pin = self.settings.second_pin
+        edge_width = self.settings.edge_width
 
         if self.rightside:
             spring = self.settings.spring in ("right", "both")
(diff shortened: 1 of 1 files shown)


florianfesti:
Thanks for all the work cleaning this up!

PR #160, opened 2019-08-04
New generator DividerTray

gcollic:
Thanks for this great project !

Here is a tray with dividers.
It consists of a bottomless tray, with columns (sx: as usual), and slots in the sides (sy: instead of rows) to be able to insert dividers. It also generates a single divider for each column.


gcollic:
Actually WIP: I think I will add some options, and a picture. Feel free to give some feedback :-)

florianfesti:
I wonder if the middle walls could be made less tall. There is no need for them to go down to the bottom but only need to be strong enough to carry the dividers.

The saved material could be used to doubled them up in width. That way you can set the dividers everywhere even side by side.

A possible extension would be using a bottom_edge parameter to get a option for a bottom. If you want to be able to move the tray with cards/papers in it that may become handy.

I would probably hard code the slot_radius to something like 1 thickness. Actually most of the divider features could be done that way. But that's a matter of how much control (and options to understand) you want to give to the users.

An option could be dividers angling forward.

gcollic:
Thanks for the feedback!

A little more context about what I'm going for: my main use case is creating game insert, where you can have more games/extensions inside the main cardbox than with the original insert (3-in-1 like Dominion, or even 7-in-1 like Unlock). Plywood dividers are too fat most of the time to separate each of the cards types (cumbersome and too much space taken), they are separated with thin paper/plastic dividers instead. The few plywood dividers are more like "movable walls", to separate whole extension from another, or to keep the card from already-full spaces falling into empty spaces (extensions you don't already have, etc).

This use-case uses so few dividers that there is no need for doubled-up walls, but I definitely see the appeal to it for other use cases, I will try to look into it. We could also have an option for the divider's tab to only use half the thickness of the wall (both have differents advantages). 

The bottom-edge has value too, I tried to keep the complexity on the divider part for a start, but I agree it could be a great addition.

For the settings, I prefer to err on the side of too much settings than too much hard-coded for a start, and then removing some if needed. I would love to categorized my arguments in two: core and advanced, and only show the core one by default but accepting the advanced one when given (and showing them with a parameter in CLI or a "advanced settings" button in the web interface). Is there any plan for this kind of features ? 

Angling dividers were my next big improvement on that :-) I did the math to take care of the radius and such, I should add it soon.

I think there is definitely a use for "fixed" dividers (especially angled one), and movable ones. I will try to support both needs (with/without extra slacks, etc).

gcollic:
I added :

- Option to not have some walls on slides (when the main cardbox of the game have some place left for other things)
- Option to add slack in divider slot to easily insert and remove them
- Fix : removed height adjustement when using --outside since there is no bottom right now

(I will squash them when I'm done)

florianfesti:
One way to make the settings nicer would be to create a Settings class for the dividers similar to the Edge settings. That way you could have a DividerSettings tab in the web interface. This also provides a convenient way of having features relative to the thickness.

florianfesti:
If you support angled dividers you may want to offer an option to have one at the very end (or angle the back wall, too). This allows for all cards being at the same angle.

florianfesti:
Just to be clear. All these ideas and improvements are optional. The generator is fine as it is right now and can be merged if you want. It is up to you to decide when it is finished. Some features can also be added later. You can also leave some features like adding the bottom_edge param to me if you don't wont to bother with them. Otoh if you are working on these features anyway it is better to finish them before merging the patches.

gcollic:
I added the angle settings. I plan to add another setting to propose to compute the tray height based on the size of the card and the angle at which they are.

I restore the context and move manually after tracing the slotted edge, in order to fix some rounding errors which were accumulating. It was a small drift, but I think it's safer this way.

Would this edge be usefull with the main ones, or just for this box ?

gcollic:
@florianfesti I pushed some new features, and a picture.

The code is a bit more generic than what the settings uses actually. It enables to do more complex box insert by hand (such as mixing multiple angles, like some game inserts needs) and it could help to reuse the dividers edge in other boxes later (such as TrayLayout for example, it could be great).

The parameter `--outside` is very similar to other boxes, but the angle increases the differences between outside and inside measures. When the measures are `inside`, I suppose the user gives all sizes according to the content which needs to fit between the dividers. Even the height of the tray is computed so that the height of the cards between the dividers are equals to `--h` (and not the tray itself in this case)

> If you support angled dividers you may want to offer an option to have one at the very end (or angle the back wall, too). This allows for all cards being at the same angle.

I implemented this with `sy`. If the user ask for `300/3`, he only gets the 2 dividers between the 3 cells (like inner walls in other boxes), but if he asks for `0:300/3:0` he also gets a divider at the start and one at the very end. It seems logical to me. Is it OK for you ?

For the last cell (or divider) to be useful, I also enlarge the box's length, so that the wall's bottom makes way for the cards in spite of their angle.

I also included debug information in the generated SVG (in case of `--debug`), to help find the right parameters when needing precise size both outside and inside (such as having the tray has the exact same size as the game box, but also dividers which have the right size for each deck of cards).

I might add the angled back wall too, bottom edge, or the smaller and doubled up slot's edge, but maybe not now.

I am ready for the merge.
What do you think? I can change things if needed.

gcollic:
It was not compatible with python2, it is now

florianfesti:
Ok, finally got the time to have a deeper look. I still have the feeling things could be simplified. But I just merge it for now and see where it leads to in the future.

Thanks for your work!

PR #161, opened 2019-08-12
Add picture for DiscRack

chrysn:
This image is a bit unorthodox in that it shows the assembled rack with actual discs inserted. I think this is suitable here as without them, it looks like a bit like a scaled-down park bench, and the intended usage pattern may not be immediately obvious.


florianfesti:
Thanks!

PR #162, opened 2019-08-12
add h19rack for a ported half 19inch rackmount

foosinn:
useful to place ports next to a half size rack thing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.ud
 build/
 dist/
+env/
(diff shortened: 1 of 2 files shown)


florianfesti:
Pushed. Renamed generator and made some minor adjustments. See c3185f51295ebf6ab9d63942a824cc9ddd091e53

Thanks for the patch!

florianfesti:
An image for the generator would be a great addition...

PR #176, opened 2019-10-04
New Generator : MicroRack

gmorell:
Thanks for the library, 

I've created a parametric rack for single board computers. It consists of stacked levels and sidewalls with cutouts for the USB/network ports

Images.
https://i.gng.io/media/8bf4b7e86cf04bdfc5ab29ad54bf5547.jpg
https://i.gng.io/media/5057647ebb47130ccb6dd2b34587540c.jpg


florianfesti:
Thanks for the PR. There are some issues with the parameters when using the web interface as it does not support decimal.Decimal or dashes in the parameter names. But I already have a patch for that I can just apply before merging. I should really add a check that rejects unsupported parameter types...

Other than that this looks pretty neat. I wonder if it is possible to reduce the number of parameters a bit. May be some clearance can be just added to the size of the board itself? I also wonder if adding a set of preselected mounting patterns to choose from might be more convenient for users. Not sure if having several boards to choose from is really better.

As the feet are not really needed for airflow you could add a top_edge and bottom_edge parameter (yes, I just complaint about too many parameters...)

florianfesti:
Oh, can you please add a picture to your PR. There are too many generators already without.

florianfesti:
While I like the generator for being simple I was wondering whether it would be even better if one could remove the panels with the computers on them more easily. Right now you can only unscrew them from the side which seems pretty inconvenient. I already have some rough ideas. Is this something you would be interested in trying out?

florianfesti:
Pushed upstream as 569eb2f49c601e0d96b68c894293c66685a381fb
The fix for the parameter names is merged right in. Some other fixes I left as separate patches. I finally added one of the liked images as a thumbnail. I hope that's fine with you.

Thanks for submitting the PR!

PR #181, opened 2019-10-23
Generate thumbnails statically

lujoga:
Currently, the preview images shown for some of the generators are full-size pictures scaled down by the browser. To improve load times, 200px-wide thumbnails should be generated using the `scripts/gen_thumbnails.sh` script (has to be executed from within the `scripts` directory). This is done automatically when the Docker container is built. The `boxesserver` has been adjusted to serve those thumbnails instead of the original pictures.

Fixes #68


scripts/gen_thumbnails.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -e
+
+STATIC_DIR=../static/samples/
+THUMB_WIDTH=200
+THUMB_HEIGHT=10000 # height: auto;
+
+rm -f $STATIC_DIR*-thumb.jpg
+
+for f in $STATIC_DIR*.jpg
```
Rotzbua: Better:

> for f in "$STATIC_DIR"*.jpg

I guess the same for `rm -f`
Source:
```
$ shellcheck myscript
 
Line 11:
for f in $STATIC_DIR*.jpg
         ^-- SC2231: Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .

lujoga: Fixed it, thanks. For me, `shellcheck` now returns without any complaints.

florianfesti:
Overall this looks like a good solution. I would probably prefer if the thumbnails would be part of the repository. For this the thumbnails should probably not be overwritten if they already exist, to not add more changes to the repo than necessary.
Generating them in the Dockerfile is not really a solution as Boxes.py can be deployed in various different ways. In theory they could be generated in the CI but that's probably a bit overkill.

lujoga:
The generated thumbnails are now part of the repository, being regenerated only when the original file changes. The `Dockerfile` is restored to its previous state.
I have also modified the HTML template and style, so the picture is shown relative to the viewport (i.e. `position: fixed;` in CSS). Previously, it has been positioned at the top of the section the respective generator belongs to, which sometimes is off-screen.

florianfesti:
Thanks for taking care of this.

PR #182, opened 2019-10-31
new FR translation

georgesk:
Hello, here is a French translation


florianfesti:
Thanks for all the work that must have gone into this!

PR #184, opened 2019-10-31
modification to allow one to package boxes for Debian

georgesk:
Hello, I am beginning to create a debian package for Boxes.py. You can have a look at https://salsa.debian.org/georgesk/boxes

In order to make easier the packaging, and lower the count of debian patches, I would like to propose you to accept the present PR modifying setup.py

The modification should have no action on ordinary installations; however upon debian packaging, when the command fakeroot is called to embed the packaging process in a new environment, setup.py behaves differently in order to comply with standard file hierarchy.


georgesk:
hmmmm.... this pull request is not mature enough, I must close it now

PR #185, opened 2019-11-01
Setup for debian

georgesk:
Hello, now the PR is mature.

When setup.py is run inside a packaging routine, one has to define an environment variable named
CURRENTLY_PACKAGING, and a special behaviour is used by setup.py to comply with the
File Hierarchy Standard.

This patch does not modify setup.py's behavior for ordinary builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,9 +46,8 @@ class CustomBuildExtCommand(build_py):
         self.execute(self.generate_mo_files, ())
         self.execute(self.buildInkscapeExt, ())
 
-        if 'FAKEROOTKEY' in os.environ:
-            # we are probably running under fakeroot
-            # so we are probably building a Debian package
+        if 'CURRENTLY_PACKAGING' in os.environ:
+            # we are most probably building a Debian package
             # let us define a simple path!
             path="/usr/share/inkscape/extensions"
             self.distribution.data_files.append(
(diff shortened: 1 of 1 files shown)


georgesk:
hmmmm... about the rationale of this PR:

I am trying to turn Boxes.py into a Debian package, to be introduced in Debian repositories (hence Ubuntu and other derivatives). The work in progress can be seen at https://salsa.debian.org/georgesk/boxes and the debian package is now correctly built by a Continuous Integration pipeline.

setup.py
```diff
@@ -45,17 +45,30 @@ def run(self):
...
-            pass # Inkscape is not installed
+        else:
+            # we are surely not building a Debian package
+            # then here is the default behavior:
+            try:
+                path = check_output(["inkscape", "-x"]).decode().strip()
+                if not os.access(path, os.W_OK): # Can we install globaly
+                    # Not tested on Windows and Mac
+                    path = os.path.expanduser("~/.config/inkscape/extensions")
+                self.distribution.data_files.append(
```
florianfesti: Guess adding the files to the self.distribution does not really need to be in the try except block. So we could have only one place in the code that does that. Probably by setting path to None and using another if block.

florianfesti: Never mind I'll just merge as is and fix that later.

PR #188, opened 2019-11-13
Windows/pyInstaller compatibility

deepinthewoods:
This fixes an OSError when running the inkscape extension on windows.


deepinthewoods:
So, pkgutil.walk_packages() does not work within a pyInstaller exe file, even if they are added to the --hidden-imports. I have made a workaround which only happens if the list of packages is empty i.e it is running as an exe/inkscape extension. It relies on a pre-generated list of the generator names.

florianfesti:
Great that someone looks into the Windows installation! Thanks a lot!
But to be honest I am not a big fan of the proposed patches. I understand that they are working around serious issues and if they are the only way to get things working I will happily 
add them. But may be we can have a look at each issue and see if we can come up with a more elegant solution.

florianfesti:
This write permission thing seems kinda odd. Inkscape should pass a proper filename to the boxes.exe. If this is about using `boxes.exe` on the command line I see two options:
1. Have the default output point to a location in the user dir
2. offer a unix style ~ for the home/user directory
or do both at the same time.

florianfesti:
While rewriting the inx files is a good quick fix I would rather generate them properly. It is easy to adjust boxes2inx to do that. But there is bit wider discussion to be had:
There are two options to call boxes.py form Inkscape: As a plain executable (which works very well on Linux) or as a Python script with 
`<command reldir="extensions" interpreter="python">boxes</command>`
I have avoided the later as there may be multiple Python installations and I did not want to rely on boxes being installed in the same one used by Inkscape. Not sure how this pans out on Windows.

So if the executable on Windows is always called boxes.exe and that's what Inkscape should use we can just change boxes2ink to generate the files that way on Windows.

Another question is whether the excutable really should reside in the Inkscape extensions directory but I guess that's the easiest solution for now.

florianfesti:
The work around for pkgutil.walk_packages is obviously the ugliest part but also the one where I have the most difficulties on suggesting another solution. Can you comment on why pkgutil.walk_packages does not work in the confine of pyInstaller? Can't we just add boxes/generators/*.py without importing anything?

deepinthewoods:
The write permission is because inkscape is executing boxes.exe from the
base inkscape folder. So it's trying to write to inkscape/box.svg instead
of extensions/box.svg.
Just changing to default location to the home dir would fix it, yes.

deepinthewoods:
The user can specify a python interpreter by editing inkscape's
preferences.xml if they're running from source, in which case the INX files
need to be as you've written.
So both options would be useful/necessary.

deepinthewoods:
The pkgutil thing is just from pyInstaller, you should be able to reproduce
it in linux. I'm just running 'pyinstaller -F boxes' from the
scripts folder. The resulting executable in the scripts/dist folder gives a
"generator not found" error when trying to generate any box.

pyInstaller has a --hidden-imports param which would seem to be useful but
I haven't been able to get that to work. I was generating a hooks file with
the hiddenImports in there originally.

I found this issue discussing pkgutil.itermodules
https://github.com/pyinstaller/pyinstaller/issues/1905#issuecomment-525221546

Maybe you could just replace walk_packages with something else that
pyInstaller does support? The way I've done it is terrible but it works
so ¯\_(ツ)_/¯ I could probably keep my fork up to date easily since it's
just the one small py file, so it's not a big deal if you don't even fix
this one.

florianfesti:
Ooof, this is getting old.  Closing. The whole setup.py business needs more love but this is not the direction I want to go.

PR #189, opened 2019-11-20
Add parameters to makita power supply generator

mbrgm:


florianfesti:
There's still more to do here but this is a good first step of cleaning things up. Thanks!

PR #199, opened 2020-01-05
More representative sample image for DividerTray

gcollic:
Hello,

This picture is more explanatory of what DividerTray is able to achieve than my previous one (with angled slots, etc.).

Thank you


florianfesti:
Added updated thumbnail and pushed. Thanks!

PR #201, opened 2020-01-22
fixed d3 diameter (mounting screw holes)

stif:
In the generator file for electronicsbox.py was a typo. it should be d3 instead of d2..


florianfesti:
Thanks for the patch!

PR #204, opened 2020-02-16
Add a bottom to the divider tray.

pokowaka:
This add support for the generation of a bottom for the divider tray.
The bottom can be generated by passing the --bottom=True flag.

This makes it possible to generate ./static/samples/DividerTray.jpg

### DIFF
--- a/boxes/generators/dividertray.py
+++ b/boxes/generators/dividertray.py
@@ -17,7 +17,7 @@
 
 from boxes import Boxes, edges, boolarg
 import math
-
+from functools import partial
 
 class DividerTray(Boxes):
     """Divider tray - rows and dividers"""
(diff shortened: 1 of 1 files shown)


boxes/generators/dividertray.py
```diff
@@ -130,6 +137,11 @@ def render(self):
             max(facing_wall_length, side_wall_length), self.h, "ffff", move="right only"
         )
 
+        # Bottom piece.
+        if self.bottom:
+            self.rectangularWall(facing_wall_length, side_wall_length, "ffff",
+                    callback=[partial(self.generate_finger_holes, side_wall_length)], move="up")
```
florianfesti: This looks a bit fishy. Shouldn't the edge type be "e" for the sides that don't have a side wall attached? As happens when left_wall or right_wall is False?

pokowaka: It looks like you might be right :-).

florianfesti:
Ok, fixed the side edges of the bottom and commited as a0338a559f55afa260b3beb1f7f0d3ea1caa55b5

florianfesti:
Oh, thanks for the patch! There are several people eager to use this new feature:
https://hackaday.io/project/10649-boxespy/discussion-139426

PR #218, opened 2020-04-10
laserguide (wip)

foosinn:
The basic idear for a laserguide that is held in place by a few fingerjoints.

Not ready to merge since its neither parametric nor tested for correct measurements. Just a PR to document that there is some work done.

Was made for the Backspace faceshields and works for that :D


florianfesti:
I still have this in some local branch and may finish it at some point. But it does not look like keeping this PR open is helping in any way. Closing for now to keep the back log clean.

PR #232, opened 2020-07-29
add requirements file

njhurst:
Rather than interpreting the readme, if we include a requirements.txt file we can automate the dependencies.


florianfesti:
Good idea! Thanks for the patch!

PR #236, opened 2020-08-14
fixed bug where the option hexpattern=True PaintStorage would cause an error.

Ba-So:
The issue was caused by an erroneous variable being passed.
This happened since hexHolesRectangle was expecting a settings object but got a tuple passed.


florianfesti:
Yeah, looks like I missed that one. Thanks for the fix!

PR #237, opened 2020-08-14
corrected use of settings. paintbox.py works properly now

Ba-So:


florianfesti:
Merged manually. Thanks for the additional fix!

PR #238, opened 2020-08-26
Add "Save to URL" button

MarcinZukowski:
This PR adds a button which modifies the current URL to include the current settings. 
This allows e.g. bookmarking or sharing designs _before_ they are rendered, allowing modifications etc.


florianfesti:
This is a great idea. Thanks for your work!

PR #240, opened 2020-08-28
Adding "Grooved Edge"

MarcinZukowski:
This adds an edge type that contains grooves, with differing styles/widths/counts etc.
This could also replace the GrippingEdge I guess.

Wasn't sure if this should be a separate edge type, or perhaps a modification to the Streight Edge. I chose to have a separate edge type to allow per-edge determination if grooves are desired. 

Added an example of how this could work in UnevenHeightBox. Not sure if this is the right approach frankly, but it works nicely there.

Some interesting (IMO) examples of what this can produce:

```
scripts/boxes UnevenHeightBox --edge_types zeee --x 60 --y 60 --lid=1 --lid_height=20 --height0=40 --height1=40 --height2=50 --height3=50 --reference=0

scripts/boxes UnevenHeightBox --Grooved_gap=0.2 --Grooved_arc_angle 90 --Grooved_width 0.2 --Grooved_margin=0.2 --Grooved_style=softarc --reference 0 --edge_types zzez --x 60 --y 60 --lid=1 --lid_height 10 --height0=30 --height1=30 --height2=40 --height3=40

scripts/boxes UnevenHeightBox --Grooved_interleave 1 --Grooved_style triangle --Grooved_gap 0.2 --Grooved_tri_angle 45 --Grooved_width 0.1 --Grooved_margin 0.0 --edge_types zzzz --x 60 --y 60 --lid=1 --lid_height=20 --outside=0 --height0=40 --height1=30 --height2=45 --height3=40 --reference=0

scripts/boxes UnevenHeightBox --Grooved_style triangle --Grooved_gap 0.0 --Grooved_tri_angle 30 --Grooved_width 0.1 --Grooved_margin 0.0 --edge_types zzzz --x 60 --y 60 --lid=1 --lid_height=20 --outside=0 --height0=40 --height1=30 --height2=45 --height3=40 --reference=0
```


MarcinZukowski:
@florianfesti Any thoughts about this one? :)

florianfesti:
Sorry, I should have replied earlier.

While I am not too keen on using up the last free letters for new edge types I really like this one. Especially as there are several more use cases for it. There are boxes where it could be used as grip hole like in the TwoPiece box or even the CardBox.

But I am a bit confused by the code. There still seem to be some issues:

http://localhost:8000/UnevenHeightBox?FingerJoint_angle=90.0&FingerJoint_style=rectangular&FingerJoint_surroundingspaces=2.0&FingerJoint_edge_width=1.0&FingerJoint_finger=2.0&FingerJoint_play=0.0&FingerJoint_space=2.0&FingerJoint_width=1.0&Grooved_arc_angle=120&Grooved_gap=0.1&Grooved_interleave=0&Grooved_interleave=1&Grooved_inverse=0&Grooved_margin=0.3&Grooved_style=arc&Grooved_tri_angle=30&Grooved_width=0.7&bottom_edge=F&x=300&y=300&outside=0&outside=1&height0=50&height1=50&height2=100&height3=100&lid=0&lid=1&lid_height=0&edge_types=ezeZ&thickness=3.0&format=svg&tabs=0.0&debug=0&reference=100&burn=0.1&render=1

There still are also some minor issues to be resolved before merging: 

Please rebase and squash the fix up patches. This should be two clean patches: One adding the new Edge and one adding it to the UnevenBox generator (Possibly more for the boxes mentioned above).

The change at the UnevenBox may lead to parts overlapping. I am not quite sure what's the right way to solve this. The technically correct way would move the lid parts orthogonal to the slopes. But that requires a lot of calculation and also spacing out the bottom pieces differently. May be there's a simplified version that's good enough.

The GroovedEdgeCounterPart needs its `margin` method implemented.

You should call `self.edge` with `(margin, tabs=1)` both times to get 2 tabs added to the edge.

MarcinZukowski:
@florianfesti I frankly don't know how to influence the overlap. I tried implementing `margin` and `spacing` in `GroovedEdgeBase` but it's never called. In general, I think it should be relatively easy to compute e.g. maximum "sticking out" of pieces from an edge. This should prevent overlap, but might cause more space waste.

Btw, there are other generators that generate overlapping pieces - I assume people know how to edit e.g. SVG before sending to laser :)

About two patches, I think that's an overkill for something like this, but if you insist, sure. Do you first want to handle the margins/overlap?

florianfesti:
> @florianfesti I frankly don't know how to influence the overlap. I tried implementing `margin` and `spacing` in `GroovedEdgeBase` but it's never called. In general, I think it should be relatively easy to compute e.g. maximum "sticking out" of pieces from an edge. This should prevent overlap, but might cause more space waste.

Sorry, I should have been more clear with this. Not having `margin` implemented and the spacing in `UnevenBox` are two separate issues. The lid parts in UnevenBox are placed "by hand" using the `.moveTo()` call. To place them properly one would probably need to draw the lid parts together with the bottom part and then move right to take the extra space needed into account.

Most other generators use only the `move` parameter to layout things. There `.margin()` and `.spacing()` is used to determine how much space is needed between parts. 

> Btw, there are other generators that generate overlapping pieces - I assume people know how to edit e.g. SVG before sending to laser :)
Yes, Boxes.py has a rather liberal attitude towards checking user inputs. Still having less overlapping parts i better than having more. I first though moving the lid row a little bit up if a GroovedEdge is selected is good enough. But it turns out this makes things worse for steep parts. May be we leave that issue alone for now and get the patch merged in first. I can take care of the parts placement afterwards - if we really think it is something worth fixing.

florianfesti:
> And let me know what part of code confuses you.

OK, I think I know understand how the different parameters are working together. But their descriptions need to be a bit better (yes, I am really not the one that should be complaining about that). Here some suggestions:

- interleave: Alternate direction of groves
- margin: minimal space left and right without grooves (fraction of the edge length)
- width:  width of each groove (fraction of the edge length)

Technically the interleave feature breaks the assumptions that all "normal" edges are symmetrical. But as all pieces need to be drawn in a counter clockwise fashion this should not be an issue except some weird edge cases (no pun intended).

MarcinZukowski:
I created #246 as a follow up, will close this one

PR #243, opened 2020-09-18
Agricola Insert

gcollic:
A full insert for the Agricola game (Revised edition), specifically designed for the "Farmer Of The Moor" expansion, and all the sizes also take into account the 5-6 player expansion pieces (but not actually tested since I don't have it yet).

I refactored my DividerTray box in order to reuse the divider edge for a similar but different box. I also used polyline since I discovered it (and Black did change some indentation).
The divider edge (with its SlotDesciption utilities) could be moved with the other edges if you prefer ?

I also added a not-centering option on rectangularHole.


florianfesti:
Wow! This is an impressive piece of work.

I have not looked too deeply at all the code but at first glance things look pretty good. So far there are only a few minor nitpicks:

- Please divide this into 3 separate patches (basically by file). Keeping it as one PR is fine.
- The feet of the simple_tray_divider could be arranged a bit more space saving by tilting them 45°.
- The card_divider_tray should probably have 2 or even 3 wooden dividers rendered - or a description telling people to cut the number they need/want
- The paper divider need to be separated and/or needs some explaining text
  - either in the `description` attribute that is shown at the bottom of the page
  - or on the drawing itself
- The `description` can also have some assembly instructions. E.g. the simple_tray_dividers are not 100% obvious from the pictures alone - even if they are very simple.

gcollic:
I took your feedback into account, described more the assembly, and added a more explicit description :-)
The simple divider feet are now rendered 2 by 2, in a more optimized way.
Let me know if you have any more feedback :-)

boxes/generators/agricolainsert.py
```diff
@@ -0,0 +1,882 @@
...
+Here are the different elements, from left to right in the generated file.
+
+#### Card tray
+
+The cards are all kept in a tray, with paper dividers to sort them easily. When
+the tray is not full of cards, wood dividers slides in slots in order to keep
+the cards from falling into the empty space.
+
+There should be enough space for the main game, Farmers Of The Moor, and the 5-6
+player expansion, but not much than that.
```
florianfesti, line 56: but not much *more* than that.

florianfesti:
I am not quite sure what I think about the use of `.move()` in `.render_simple_tray_divider`. That's not quite how it was intended to be used. Otoh it seems to work. Still, using `.moveTo()` might be more in line with the rest of the code base. But that's nothing stopping the PR from being merged.

With the wording issue above fixed this is ready for merging IMHO.

boxes/generators/agricolainsert.py
```diff
@@ -0,0 +1,882 @@
...
+            ],
+        )
+        for _ in range(2):
+            self.rectangularWall(
+                support_length, bed_foot_height - self.thickness, "feee", move="up"
+            )
+
+        self.ctx.restore()
+        self.rectangularWall(
+            bed_inner_length, bed_inner_width, "FFFF", move="right only"
```
florianfesti, line 585: This shoul be "FeFF" to close the gaps between the columns

boxes/generators/agricolainsert.py
```diff
@@ -0,0 +1,882 @@
...
+        """
+        self.ctx.save()
+        bed_inner_height = player_box_height - self.thickness
+        bed_inner_length = 66.75
+        bed_inner_width = 50.5
+        bed_foot_height = 6.5
+
+        bed_edge = Bed2SidesEdge(self, bed_inner_length)
+        noop_edge = NoopEdge(self)
+        for _ in range(2):
```
florianfesti, line 545: With a bit of moveTo() magic these could probably be placed a bit better by turning the second part 180°.

boxes/generators/agricolainsert.py
```diff
@@ -0,0 +1,882 @@
...
+                partial(
+                    self.generate_bed_holes,
+                    bed_inner_width,
+                    bed_hole_margin,
+                    bed_hole_length,
+                    support_length,
+                )
+            ],
+        )
+        for _ in range(2):
```
florianfesti, line 578: May be these can fit right of the wall with the BedHeadEdge when turned 90°.

florianfesti:
There is also an issue with the rendering of the bullet points in the `description`. This is an issue with the CSS and not your fault. Please ignore for now. I'll try to fix that soonish.

boxes/generators/agricolainsert.py
gcollic: I totally agree, in fact I actually cut them like you are suggesting :-) But most of the time I don't use boxespy for layout optimization. I prefer to optimize more aggressively directly in Inkscape, and also take into account the shape of the piece of plywood I use.
I find that using moveTo adds complexity in this case, I did a separate commit for that.

gcollic:
> I am not quite sure what I think about the use of `.move()` in `.render_simple_tray_divider`. That's not quite how it was intended to be used. Otoh it seems to work. Still, using `.moveTo()` might be more in line with the rest of the code base.

`moveTo` doesn't create a new `path`, and afterwards it's difficult to manipulate elements which were created inside the same `path`. I optimized this code a bit more, there is still a `move` for each piece, but in a way which seems more in line with the rest.

gcollic:
I took your feedback into account, feel free to add more if you have any :-)

gcollic:
(I marked the comment I fixed as resolved to signal it was taken into account in my last push, but I probably should have left them open for you to resolve instead)

florianfesti:
> `moveTo` doesn't create a new `path`, and afterwards it's difficult to manipulate elements which were created inside the same `path`. I optimized this code a bit more, there is still a `move` for each piece, but in a way which seems more in line with the rest.

Yes, there's some point to that. An alternative could have been using `polygonWall` - for you I still need to write a doc string and add it to the docs. 

> (I marked the comment I fixed as resolved to signal it was taken into account in my last push, but I probably should have left them open for you to resolve instead)
No big deal. Yes, this is how GitHub thinks they should be used. No reason for us for feeling bad about doing it otherwise.

Anyway, let's not optimize this for ever. This already is one of the most polished generators.

florianfesti:
This deserves an entry in the project log of the Hackaday.io page: https://hackaday.io/project/10649/logs
You are welcome to write that yourself. Otherwise I am also happy to write it. Feel free to apply for membership in the Hackaday.io project if you want.

florianfesti:
@gcollic I guess the insert only fits the box when cut in 3mm material (or less). May be this should be mentioned in the first sentence of the description or at least the first after the introductory  paragraph.
I can add that if you confirm the "approved" thickness range.

florianfesti:
For the record: https://hackaday.io/project/10649/log/185370-its-all-fun-and-games

PR #249, opened 2020-10-02
Add generators for marble run parts

lujoga:
### MarbleCatcher

An arbitrarily-sized box with an open top and a customizable number of clips, which (together with `MarbleSupport` parts) are used to attach it to a breadboard.

### MarbleSlide

A marble slide with a notch in one side, so it can be "hung" on one `MarbleSupport` and stabilized with another opposite to the notch. Supports slopes from horizontal to 100%.

### MarbleSupport

The aforementioned support part for mounting other parts to a board of evenly spaced holes (default spacing is 15mm).

### DIFF
--- a/boxes/generators/marblecatcher.py
+++ b/boxes/generators/marblecatcher.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+# Copyright (C) 2020  Luca Schmid
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


florianfesti:
With this project looking stalled am closing this here for now. I'd really like to have a marble run in Boxes.py but this one requires a specific peg board and though is a bit specific. With at least a generator a peg board (stand) and a perspective of more marble obstacles being added I'd be happy the merge this -  probably into its own group right away.

So feel free to re-open when this goes forward again.

PR #254, opened 2020-11-01
Add platonic solid generator

not7cd:
Rough version with no settings exposed.
I had to create a custom finger generator, to unevenly place leftover space.
I'm generating only positive fingers. I have worry, that they will won't fit together, and negative fingers account for some play, that I'm not sure about.

Inspiration.
DIY 3D Platonic Solids Set  Sacred Geometry  Polygon Set  image 0

Here I compare two generated sides.
Screenshot from 2020-11-01 16-59-09

### DIFF
--- a/boxes/generators/platonic.py
+++ b/boxes/generators/platonic.py
@@ -99,7 +99,7 @@ class Platonic(Boxes):
     def __init__(self):
         Boxes.__init__(self)
 
-        self.addSettingsArgs(edges.FingerJointSettings)
+        self.addSettingsArgs(edges.FingerJointSettings, surroundingspaces=0)
 
         self.buildArgParser(x=60, outside=True)  # x should be treated as edge length, TODO: change that
         self.argparser.add_argument(
--- a/boxes/generators/platonic.py
+++ b/boxes/generators/platonic.py
@@ -108,11 +108,6 @@ class Platonic(Boxes):
             help="type of platonic solid")
 
 
-    def regularTrianagleWall(self, a, edges="uuuu", move=None):
-        self.regularPolygonWall(3, side=a, edges=edges, move=move)
-
-
-
     def render(self):
         # adjust to the variables you want in the local scope
         e = self.x
--- a/boxes/generators/platonic.py
+++ b/boxes/generators/platonic.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright (C) 2013-2016 Florian Festi
+# Copyright (C) 2020 Norbert Szulc
 #
 #   This program is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
--- a/boxes/generators/platonic.py
+++ b/boxes/generators/platonic.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+# Copyright (C) 2013-2016 Florian Festi
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from boxes import *
+from boxes.edges import FingerJointBase, FingerJointEdge
+
+from math import sin, pi
+
+class UnevenJointBase(FingerJointBase):
+    def calcFingers(self, length, bedBolts):
+        space, finger = self.settings.space, self.settings.finger
+        fingers = int((length - (self.settings.surroundingspaces - 1) * space) //
+                      (space + finger))
+        if not finger:
+            fingers = 0
+        if bedBolts:
+            fingers = bedBolts.numFingers(fingers)
+        leftover = length - fingers * (space + finger)  # we don't another space
+
+        if fingers <= 0:
+            fingers = 0
+            leftover = length
+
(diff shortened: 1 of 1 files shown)


not7cd:
I also don't understand what bolts do, as I don't see an obvious connections in settings in other generators.

Blancanegra:
Wow, those are beatiful! Can't wait for the final generator, thank you!

florianfesti:
> I also don't understand what bolts do, as I don't see an obvious connections in settings in other generators.

These produce slots for nuts and bolts but are not well integrated into the code yet. You can just ignore them. Just delete the traces from the code.

not7cd:
I hoped someone would comment on my code. But if you are amazed by finished product, please see original creation:https://www.etsy.com/listing/526639305/diy-3d-platonic-solids-set-sacred

boxes/generators/platonic.py
```diff
@@ -0,0 +1,174 @@
...
+
+from boxes import *
+from boxes.edges import FingerJointBase, FingerJointEdge
+
+from math import sin, pi
+
+class UnevenJointBase(FingerJointBase):
+    def calcFingers(self, length, bedBolts):
+        space, finger = self.settings.space, self.settings.finger
+        fingers = int((length - (self.settings.surroundingspaces - 1) * space) //
```
florianfesti, line 25: Using surroundingspaces doesn't really make sense for this kind of edge. It allows the FingerJointEdge to squeeze one finger in less length than one finger and two spaces. Unless you have a the leftovers on a different height this doesn't make any sense.

boxes/generators/platonic.py
```diff
@@ -0,0 +1,174 @@
...
+                      (space + finger))
+        if not finger:
+            fingers = 0
+        if bedBolts:
```
florianfesti, line 29: bedbolts don't really work for non 90° joints. I suggest you just scrap that code paths.

boxes/generators/platonic.py
```diff
@@ -0,0 +1,174 @@
...
+                self.polyline(
+                    0, -90 * p, 0.8*h, (90 * p, 0.2*h),
+                    0.1 * h, 90, 0.9*h, -180, 0.9*h, 90,
+                    f - 0.6*h,
+                    90, 0.9*h, -180, 0.9*h, 90, 0.1*h,
+                (90 * p, 0.2 *h), 0.8*h, -90 * p)
+            else:
+                self.polyline(0, -90 * p, h, 90 * p, f, 90 * p, h, -90 * p)
+
+        self.edge(0, tabs=0)  # Whole point of this class
```
florianfesti, line 90: This is not doing anything. Which means you are also loosing the second tab on this edge. But that should not be a big issue and also can be added later to the last line of the last polyline() above.

boxes/generators/platonic.py
```diff
@@ -0,0 +1,174 @@
...
+
+        s = edges.FingerJointSettings(self.thickness, relative=True,
+                                      width=self.thickness)
+        
+        #adjust surrounding spaces, to account for finger width at edge (?)
+        # fix_r, fix_h, _ = self.regularPolygon(corners, side=1)
+        cutin_fix = s.width/sin(2*pi/corners)/s.space
+        print(cutin_fix)
+        s.surroundingspaces=0+cutin_fix
+
```
florianfesti, line 154: You should also set the `angle` attribute here. The code automatically adjusts the finger length to the angle given so all the edges from both side line up.
Note: For now this only supports joints with holes. For a joint with overhangs that can be ground away to create a closed joint other values are needed. This is a feature that should go directly into the FingerJoint Classes.

boxes/generators/platonic.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+    def render(self):
+        # adjust to the variables you want in the local scope
+        x, y, h = self.x, self.y, self.h
+        a = x  # side length
+        t = self.thickness
+
+        solid = "Octahedron"
```
florianfesti, line 142: See https://github.com/florianfesti/boxes/blob/master/boxes/generators/angledbox.py#L33 for an example on how to add a GUI parameter for such a selection.

boxes/generators/platonic.py
```diff
@@ -0,0 +1,131 @@
...
+        self.addSettingsArgs(edges.FingerJointSettings, surroundingspaces=0)
+
+        self.buildArgParser(x=60, outside=True)  # x should be treated as edge length, TODO: change that
+        self.argparser.add_argument(
+            "--type",  action="store", type=str, default=list(self.SOLIDS)[0],
+            choices=list(self.SOLIDS),
+            help="type of platonic solid")
+
+
+    def regularTrianagleWall(self, a, edges="uuuu", move=None):
```
florianfesti, line 111: This can also go way, too.

florianfesti:
This is coming along nicely and looks nearly ready for merging. One thing that still needs to be done is setting the `angle` in the FingerJointSettings. IIRC 0° is straight one while 180° is turning back on itself.

One should probably check the angles < 90° - the Tetrahedron in this case as these angles haven not really been used much, but need a different way of resolving the issue that we can't have angled cuts. They may just be fine but I am not 100% sure.

not7cd:
I still need to do some real live testing when I get to a laser cutter. About angles, I'm not so sure about them. I guess, we could get by without angle setting.

not7cd:
We are good to go
image

boxes/generators/platonic.py
not7cd: I'm still not sure how to go about it. From my real life tests everything worked fine. Fingers don't overhand and neatly finish at the solids edge.

florianfesti:
Unfortunately my hacker space is still closed. Let's just merge this as is. We can adjust the finger joints to the angle between face later on. While there is basically only one line (and the values of angles) missing I need to test this properly. I don't really trust the code to work for all angles - smaller and bigger than 90°.

Can you please squash the commits into one for merging.

florianfesti:
Oh, right! GitHub has the `Squash and Merge` button now.

Thanks for the generator!

not7cd:
I kinda forgot about it. Thanks for merging (and squashing)!

PR #258, opened 2020-11-16
Update fr.po

PaoloElMarinelo:
Minor change to make parameters easy to read and understand in french


florianfesti:
Thanks for the corrections!

Issue #274, opened 2021-01-27
DividerTray wrong box length


Feredyr:
Hello,
Thank you for this amazing work !
I try to use the DividerTray but I have an issue when I setup my box.
The box length did not fit what I setup:
![Dividerconfig](https://user-images.githubusercontent.com/78080058/105981652-9c629780-6096-11eb-961a-b73e02e867b6.JPG)

The result
![Result](https://user-images.githubusercontent.com/78080058/105981655-9cfb2e00-6096-11eb-9a66-c83109890512.jpg)

Could you please help me ?

Thanks!

gcollic:
Hello !

When the "exterior" parameter is enabled, you are telling this box what should be the total length.

But when the "exterior" parameter is disabled (like in your example), you are telling this box what should be the length of the content for each slot. The last slot ends at the right place, but requires some extra space above it, and I think it explains why it's not the size you wanted:
![image](https://user-images.githubusercontent.com/763200/105983938-bd78b780-6099-11eb-806b-dea9c2e2bee0.png)

Most of the time, you both want precise information about inner size (how much stuff each slot needs to store) and outer size (how much space I have for the entire box). I added some information in debug mode to help find the right parameters to have both the right inner and outer sizes.

Can you give the "debug" checkbox option a try and tell me if it helps ?

Don't hesitate to tell us how we could improve this configuration for clarity. "Outside" and "debug" are used by other boxes, the  text is not as helpful as it could be for this particular box, I might need to create a new parameter of the same name but a different description. Or I might just add some text to the description (it goes below the form, with the picture).

Thank you !

gcollic:
It looks like there is still an error after all. The second slot is bigger than asked and than the first. I should look into that !

Feredyr:
Thank you ! !
Enabling the exterior parameter fix almost the issues. 

 From my point of view, the text could be clarify a little bit.
if I may help:
This kind of box is mostly done for board game insert (Like in my case). So the X / Y / Z exterior parameters are mandatory to really fit the cardboard box. Then we divided the by number of section ( Sx). 
Where there can be mistake is with the Z (H) parameter.  Without the exterior checkbox, the H is the card length (in the tray). Maybe you could remove the slot angle parameter and ask for card length (as this is just math)
One other improvement than can be done, is to make start the Slot against one wall. That’s avoid space lose.
Again thank you for this amazing work

gcollic:
> start the Slot against one wall

[...]

PR #275, opened 2021-01-28
Fix DividerTray length with 'outside=false'

gcollic:
#274 was about two things: a misunderstanding, and a real bug.

This PR fixes the bug :-)

(the reporter closed the issue, I can't reopent it)


florianfesti:
That for taking care of this!

PR #280, opened 2021-02-13
Keyboards: refactor common code and add options for atreus and keypad

gcollic:
I'm planning to update [my custom keyboard](https://github.com/gcollic/grimoire60_keyboard). I'm doing some test, and obviously I'm using boxes.py this time.

My keyboard is not ready to share, but in preparation I already updated the `atreus21` and `keypad` "boxes" for better reuse, uniformity, and options.

I did the following changes:

- there was duplication
    - I created a Keyboard class to share keyboard specific functionnalities
    - keypad profits from the code structure of atreus
- `atreus`'s layout was not configurable (but the code was already parameterized), and `keypad`'s layout options were very limited
    - I created a unified arg parser (inspired by `argparseSections`)
    - the default value gives the same result as before
    - both boxes can now generate a lot of different layouts
- the values used were not documented
    - I added references to switches and socket specification sheets
    - I named some values
- there was dead code
    - I removed some code, generally copied from each other
- I fixed a slight bug in both
    - One of the layer was slightly off and assymetrical (missing a move by the length of the burn value)
    - image
- The "PCB" holes were not configurable
    - now it supports all the different Cherry MX options I know:
    - image
    - it now also supports to enable/disable the larger pin hole in case of hotswap socket
- A new switch cutout option
    - I prefer the tighter official cutout than the one used by default on these project
    - but the tighter version doesn't allow to "mod" the switch afterwards, so both options are still useful
    - image
-  The new settings are shared between keyboard project
    - with opt in / opt out

My changes are compatible with the previous version, except for the deleted `keypad.btn_x` and `keypad.btn_y` settings which were replaced by `keypad.columns_definition`. The explicit `4x3` value and the default value still give the same result as the old default `btn_x=3` and `btn_y=4`.

I would like to add a `keyboard` category some day, but right now there is still only 2 keyboard projects (one in "Box" and one in "Misc") so it's probably better to wait for a third one to have a category for them.

### DIFF
--- a/boxes/generators/atreus21.py
+++ b/boxes/generators/atreus21.py
@@ -71,20 +71,21 @@ class Atreus21(Boxes, Keyboard):
             x, y
         )
 
+    @restore
     def rim(self):
         x, y = self._case_x_y()
         self.moveTo(x * .5, y * .5)
         self.rectangularHole(0, 0, x, y, 5)
-        self.moveTo(x * -.5, y * -.5)
 
+    @restore
     def outer(self):
         x, y = self._case_x_y()
         b = self.border
         self.moveTo(0, -b)
         corner = [90, b]
         self.polyline(*([x, corner, y, corner] * 2))
-        self.moveTo(0, b)
 
+    @restore
     def half(self, hole_cb=None, reverse=False):
         if hole_cb == None:
             hole_cb = self.key
@@ -92,10 +93,8 @@ class Atreus21(Boxes, Keyboard):
         self.apply_callback_on_columns(
             hole_cb,
             self.columns_definition,
-            self.STANDARD_KEY_SPACING,
-            reverse,
+            reverse=reverse,
         )
-        self.moveTo(-self.half_btn, -self.half_btn)
 
     def support(self):
         self.configured_plate_cutout(support=True)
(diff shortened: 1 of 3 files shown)


florianfesti:
This looks really nice and is a - well actually a lot of - steps in the right direction.

As the description of the change has that many items I wonder if this should be split up into multiple patches. I know doing this afterwards is quite some pain. So I am not insisting on this. But at least the description text above needs to go into the commit message.

From a coding POV I really like that the common code is factored out and even may be re-used for more keyboard generators. I am not too keen on using static methods for this, though. The `Keyboard` class IMHO should either be an abstract super class from which the actual keyboard generators are derived from or at least a mix-in class with the keyboard generators inheriting from both `Boxes` and `Keyboard`.

gcollic:
Here are some modifications. It will be easier to review 😉

gcollic:
A better illustration of the alignment bug in keypad:
image

The exact same bug was also in atreus, and there also was an additional difference caused by a remaining magic number from keypad which was not in line with the other value anymore (4mm instead of 3.6mm, so both hands were not symmetrical)

florianfesti:
With the change split into separate patches this looks much nicer IMHO. Thanks for all the work you put into this!

PR #286, opened 2021-02-26
attempt to make burn test description more newbieproof

### LINKED ISSUE #284: Double meaning of "burn" in BurnTest, confusing for a newbie
> Set burn in the Default Settings to the lowest value to be tested. To get an idea cut a rectangle with known nominal size and measure the shrinkage due to the width of the laser cut. The burn value is half the difference of the overall size as shrinkage is occuring on both sides. You can use the reference rectangle as it is rendered without burn correction.

is the second mention of burn in "The burn value is half" referring to burn value that should be set elsewhere?

Maybe explicit "Now you can measure the burn value that you should use in other generators. It is half (...)"

matkoniecz:
fixes #284

note: I am pretty sure about the guessed meaning but it is a guess

there is a small risk that I have broken something with multiline string

How can I regenerate website and view it to check is everything OK? I looked into https://github.com/florianfesti/boxes/blob/master/CONTRIBUTING.rst and have not noticed info about it

### DIFF
--- a/boxes/generators/burntest.py
+++ b/boxes/generators/burntest.py
@@ -20,8 +20,9 @@ class BurnTest(Boxes):
     """Test different burn values"""
     description = """Set burn in the Default Settings to the lowest value
 to be tested. To get an idea cut a rectangle with known nominal size and
-measure the shrinkage due to the width of the laser cut. The burn value is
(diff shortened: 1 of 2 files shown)


florianfesti:
Thanks for the patch!

PR #290, opened 2021-03-02
try to add explanation how burn test works

### LINKED ISSUE #288: Why burn test generates multiple toothed rectangles?
https://www.festi.info/boxes.py/BurnTest

From description:

> To get an idea cut a rectangle with known nominal size and measure the shrinkage due to the width of the laser cut.

what I understand as "cut rectangle with your laser cutter, measure real width, compare with intended width to get info about shrinkage"

But burn test generates multiple pieces with joints, in multiple variants. It looks like burn test is intended to be done by checking which pieces fit (?? is it correct ??).

But then why "compare with intended width to get info" part? Maybe it describes two possible ways to measure proper burn parameter?

matkoniecz:
fixes #288

Probably can be improved further, but I am out of ideas for now.

For images: 

- do you prefer pull request or just posting image on issue tracker?
- Do you prefer to get only images of high quality or is it OK to send also mediocre ones?

### DIFF
--- a/boxes/generators/burntest.py
+++ b/boxes/generators/burntest.py
@@ -18,7 +18,10 @@ from boxes import *
 
 class BurnTest(Boxes):
     """Test different burn values"""
-    description = """Set burn in the Default Settings to the lowest value
+    description = """This generator will make shapes that you can use to select
+optimal value for burn parameter for other generators. After burning try to
+attach sides with the same value and use best fitting one on real projects.
+In this generator set burn in the Default Settings to the lowest value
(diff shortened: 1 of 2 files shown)


florianfesti:
It's still not world's greatest description but a step in the right direction. Thanks for taking care.

Any picture that is not misleading is better than no picture. Feel free to provide the picture in what ever form is convenient for you. If done by PR, please execute scripts/gen_thumbnails.sh and also add the thumbnail and the checksum file.

PR #299, opened 2021-03-17
increase play in two piece to 0.15

### LINKED ISSUE #298: Consider increasing default play in TwoPiece
Is it just me being newbie or is default play quite small? It is quite hard to open/close box. I would expect it to slide easily - and it is so narrow that box is getting damaged by opening and closing it.

PS Thank, thanks, thanks for the generator :)

matkoniecz:
note: I am very inexperienced, this is based on a single test (with 0.05 and 0.15 box) and advice from local expert
both apply to plywood

fixes #298

warning: untested

### DIFF
--- a/boxes/generators/two_piece.py
+++ b/boxes/generators/two_piece.py
@@ -30,7 +30,7 @@ class TwoPiece(Boxes):
         self.addSettingsArgs(edges.FingerJointSettings, finger=2.0, space=2.0)
 
         self.argparser.add_argument(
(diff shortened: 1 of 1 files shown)


florianfesti:
Well the fit is that tight to provide some friction. Obviously this is not a great solution. The proper fix is adding some springs to either provide friction or even snap closed. This is obviously out of the context of this PR and just increasing the play will at least make sure the pieces fit at all.
Thanks for looking into this!

PR #301, opened 2021-03-17
Add PhoneHolder

gcollic:
A smartphone holder / phone stand which holds a phone between two tabs, with access to its bottom, in order to connect a charger, headphones, and also not to obstruct the mic.

The default values are currently based on Galaxy S7.

The picture is not from the last version, I added more fingers and a back plate for a more robust build, especially if without glue, but I didn't cut another unit.


florianfesti:
Looks good. There is a placement issue with `rectangularTriangle`. While this is triggered by the generator it is not its fault.

PR #302, opened 2021-03-18
update status from beta to Production/Stable

matkoniecz:
not sure whether this classification has any impact on anything, but it seems deserving to me


florianfesti:
Well, I guess...

PR #303, opened 2021-03-18
hint for what needs to be run

matkoniecz:
at least for me it was not obvious and it seems worth documenting


florianfesti:
Yes, the docs can always be improved further. Thanks!

PR #304, opened 2021-03-20
Descriptions of ClosedBox and BasedBox

matkoniecz:


florianfesti:
Thanks for the patches!

PR #311, opened 2021-04-11
☕  Add coffee capsule holder generator

gcollic:
A simple coffee capsule holder :-) ☕


florianfesti:
Thanks for the generator!

I wonder if it makes sense to add parameter(s) for the size of the capsules. But I don't have any idea if there are different size and what they would be.

PR #314, opened 2021-04-28
Added inx parser for InkScape 1.0

vmario89:
This is a more recent inx file creator which does **not throw errors** on console (except the ones where no thumbnail files exist yet)

the old generator for 0.92 has a lot of glitches in comparison to newer Inkscape releases

grafik
grafik

inx example file:
grafik


vmario89:
Note: i made boxes.py working on both Linux and Windows. The documentation is available on https://stadtfabrikanten.org/display/IFM/Boxes.py

i made this working by having a boxes_proxy.py script which can be found at https://gitea.fablabchemnitz.de/MarioVoigt/mightyscape-1.X/src/branch/master/extensions/fablabchemnitz/boxes.py/boxes_proxy.py

florianfesti:
Sorry, for not reacting to this PR earlier. Your work on the Inkspace integration and the Windows builds is very much appreciated - especially because I don't have much smart things to say about them.

The current version of Boxes.py should no longer need cairo and should though be easier to build. It is now a pure Python project that only uses native Python modules (and the modules included in the Python standard library like gettext).

I am not sure if it really is worth keeping support for Inkscape 0.9. For me switching everything over to 1.0 is fine. But I don't have a good idea which versions are still being used.

vmario89:
hi,
thanks for updating me about cairo. Thats fine! Cairo makes it hard to deal for Windows users indeed (i use win+linux). I will try to ease up newer .exe builds when i found time for it. will update my docs so later on :)

regarding inkscape versions i follow up the opinion to drop support for older inkscape because inkscape 0.9.x uses python2. Python2 to Python3 requires a lot of workarounds - makes spaghetti coding for imports and frustrating try-except-try-except statements. Inkscape 1.0+ also deals with newer inx entities, provided by relaxng scheme. i dont see any need to use Inkscape 0.9 versions on systems personally. Users should update. It's worth to do it :-)

florianfesti:
Technically I dropped Python 2 support a while ago. So I am very much for going Inkscape 1.0 all the way.

If you are interested we can also add your instructions to the official docs (`documentation/src/` - https://florianfesti.github.io/boxes/html/install.html). As I am on Linux only I really rely on other contributors for the Windows and MacOS instructions.

vmario89:
i updated docs at https://stadtfabrikanten.org/display/IFM/Boxes.py - i removed the cairo stuff from it. Maybe you can add a link to your docs or you can just copy the complete description if you like (but they are not really clean yet)

alastaira:
Just to comment as a Windows user, I am very happy to see Cairo, Python 2.x, and Inkscape 0.9x support dropped :)

florianfesti:
Ok, let's push a bit harder to get this merged. I am all for just moving the new generator on top of the old boxes2inkscape and then just squash all changes together. I already cherry-picked the image for the Platonic Solid generator.

Is there anything that still needs fixing. May be some band aids can be pushed down into the proper place in the boxes.py code base.

vmario89:
hi,
thanks for picking up the images! =) im not aware of fixing more stuff in the inx generator yet
cheers, Mario

florianfesti:
Ok, great. I'll look into the squashing and renaming this evening.

bringert:
This change seems to break the Inkscape plugins, since it introduces a dependency on boxes_proxy.py:
https://github.com/florianfesti/boxes/commit/ecf14540f5d74d8100edcca7dce72b7e6fe26d42#diff-de66bee97ccb9f51095fed51e8266b2e39a5f90c879cc1e55298fcf3e89b2c1dR138

I can't find any code in the repo for installing a boxes_proxy.py. I worked around this by creating a symlink called boxes_proxy.py in the Inkscape extensions directory:

$ cd ~/Library/Application Support/org.inkscape.Inkscape/config/inkscape/extensions
$ ln -s ~/src/boxes/scripts/boxes boxes_proxy.py

PR #315, opened 2021-05-01
Extend PaintStorage to make stackable drawers

MrShark:
A simple extension to make "compatible" drawers that stack with the paint storage.

Add an extra argument --drawer that changes it to a simple drawer

### DIFF
--- a/boxes/generators/paintbox.py
+++ b/boxes/generators/paintbox.py
@@ -21,7 +21,7 @@ class PaintStorage(Boxes):
     """Stackable storage for hobby paint or other things"""
 
     webinterface = True
-    ui_group = "Shelf"  # see ./__init__.py for names
+    ui_group = "Shelf" # see ./__init__.py for names
 
     def __init__(self):
         Boxes.__init__(self)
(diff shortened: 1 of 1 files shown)


florianfesti:
I like this change. But the side walls should connect to the bottom with finger joints.

In general I am not a big fan if a patch does reformat the code and adds new functionality at the same time. But I know how things go. So no big deal here.

MrShark:
Added keys on the bottom.

Bit of formating ping-pong, but it hopfully evens out when I squash

florianfesti:
Hmm, after having a bit closer look: While this change keeps the width `x` the same it changes the top and bottom profile by making the inner width `x - 2 * t`. So new versions will no longer stack on top of old ones. That is not really acceptable. The variant with paint bottles need to stay unchanged and the drawers should have the same top and bottom profile and the same width.
This is a bit tricky to achieve but I will give it a try using ` ignore_widths` parameter of `rectangularWall`.

florianfesti:
Ok, that was easier than I thought. Although it is not tested yet and I really hope all pieces still fit together.

Squashed all the changes into one commit and merged.
Thanks for your work!

florianfesti:
PS: This will mean the now merged version will not stack on top of the ones you may have cut from your PR branch. Sorry!

MrShark:
Thanks and no problem, it's usually me who resolve peoples messed up git histories at work :-)

PR #316, opened 2021-05-01
Allow labeling of walls

MrShark:
Add a parameter label to rectangularWall() that
is rendered in red to identify parts, it is not ment
to be cut or etched

Added labels to the OttoBody as an example of the API

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -293,6 +293,9 @@ class Boxes:
             "--debug", action="store", type=boolarg, default=False,
             help="print surrounding boxes for some structures")
         defaultgroup.add_argument(
+            "--nolabels", action="store", type=boolarg, default=False,
+            help="Hide the labels for parts")
+        defaultgroup.add_argument(
             "--reference", action="store", type=float, default=100,
             help="print reference rectangle with given length (zero to disable)")
         defaultgroup.add_argument(
@@ -1086,7 +1089,7 @@ class Boxes:
         self.ctx.translate(*self.ctx.get_current_point())
         self.ctx.rotate(angle)
 
-    def move(self, x, y, where, before=False):
+    def move(self, x, y, where, before=False, label=""):
         """Intended to be used by parts
         where can be combinations of "up" or "down", "left" or "right", "only",
         "mirror" and "rotated"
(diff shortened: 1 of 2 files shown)


florianfesti:
This is a really great idea and a nice prove of concept. For rolling this out through the whole of Boxes.py we need to refine this a bit further.

I would probably move the actual rendering of the label move down into Boxes.move(). It knows the dimension of the parts and can make smarter decisions on where, how and whether to place the label. It would also make it easier to add labels to other parts.

Next (rather minor) thing is surrounding all the labels with `_( )` to they get picked up for translation.

At last we probably want a global switch to enable or disable the labels.

MrShark:
Thanks! I tried to implement your suggestions. 
I don't fully understand the rendering model, so there is no smarter placement of the labels :-)
I also added a label parameter to some more wall types and tested them very slightly, the placement is not perfect but it gives a hint.

I'm also thinking of adding a section something like this into the documentation:

**Colors**
The generated files uses the following color conventions:

**Black**  The outer edges of a part
**Blue**  Inner edges of a part
**Red**  Comments or help lines that are not ment to be cut or etched
**Green**   Etchings  
Normaly you will cut things in the order: Green, Blue, Black. If other colors are present, the meaning should be obvious.

I don't think there are any etchings yet but it might come, have I made the right asumptions?

MrShark:
Rebased the branch to master and squashed and split the changes to be one for the interface and one demoing the interface on Ottobody

florianfesti:
I was just about to ask...

All labels should be wrapped in `_()` though. This not only marks them for translation but it also translates them (given someone actually added a translation for the current language). So even duplicates should be wrapped.

florianfesti:
Another minor nitpick. The option should be `labels` and be `True` by default. Why making no difference functionality wise people are bad with negations and they should though be avoided in user interfaces (where possible).

florianfesti:
The drawing of the label text is not happening at the right place. This may work for parts like the rectangularWall, that ends in the bottom left corner. But there is no guarantee on where the cursor is at that point.
Move the drawing command to the end of the `if not before:` block (line 1137). There we know that the part was actually drawn and the cursor is restored to the bottom left corner by `self.ctx.restore()`.

florianfesti:
OK, just did those changes myself.

Text is now centered on the part. For very small parts we could decrease the font size. But may be I postpone this until it becomes an actual issue.

Anyway, thanks for your work. Now someone needs to label the about thousand parts in the ~120 generators there are...

PR #319, opened 2021-05-09
Fix Boxes.saved_context example in Navigation documentation

hprid:
The code example for Boxes.saved_context() uses a with statement
that does not properly call the self.saved_context method (misses
call parenthesis).


florianfesti:
OK, looks like somebody is actually reading those...

PR #321, opened 2021-05-16
Parametrizise the drillbox

MrShark:
Adds the parameters: sx, xy, h, holes, firsthole and holeincrement.

It adds two TODOs: 
One needs changes in the Boxes-class and should be in a separate pull request (changing color of the etched boxes) and one that i have not worked out how to do yet (dynamic text). I could try to include them here if you want.

My editor tends to use [black](https://black.readthedocs.io/en/stable/index.html)  to format the code as default and since I could not find any preference for what codestyle to use I kept it on all changed lines (which is more or less everything :-). If you prefer any other style, just say and I will create a new patch in that style.

### DIFF
--- a/boxes/Color.py
+++ b/boxes/Color.py
@@ -4,3 +4,9 @@ class Color:
     GREEN = [ 0.0, 1.0, 0.0 ]
     RED   = [ 1.0, 0.0, 0.0 ]
     WHITE = [ 1.0, 1.0, 1.0 ]
+
+    # TODO: Make this configurable
+    OUTER_CUT = BLACK
+    INNER_CUT = BLUE
+    ANNOTATIONS = RED
+    ETCHING = GREEN
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -72,9 +72,14 @@ def holeCol(func):
 
     @wraps(func)
     def f(self, *args, **kw):
+        if "color" in kw:
+            color = kw.pop("color")
+        else:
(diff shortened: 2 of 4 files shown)


florianfesti:
This looks pretty good already. To get another color for the rectangles you need to drawn them "by hand" e.g. using `polyline()` as the rectangularHole comes with the blue color as all holes.

Both text and colors are topics that need more work. Feel free to look into both. But merging this as is or with only slight additions is also fine. You could make the text size a parameter if you don't want to do something more complicated (right away).

MrShark:
My thinking was adding a boolean parameter _etching_ to holeCol() that defaults to False and changes the line to green (or even a color parameter, like text())

florianfesti:
I think I like the `etching` parameter better.
Color handling is on my (longer term) todo list. Basically having semantic labeling for all paths and a global settings on what colors and line widths to assign. So people can select the color scheme that works best with their tooling.

MrShark:
A new patch adding the etching parameter (and using it in drillbox)
Also documents the semantic colors

florianfesti:
Very nice. Do I look very ungrateful if I change my mind again and opt for having a `color` parameter in the `holeCol` wrapper? Now that we have semantic colors there is no need to have the semantics in the parameter name.

MrShark:
No problem, done.
Also cleaned up the commit history a bit

florianfesti:
Thanks for all the effort!

PR #323, opened 2021-05-23
fix typo

matkoniecz:


florianfesti:
Someone is actually reading the documentation. Shocking!

Thanks for the fix!

PR #328, opened 2021-06-27
Labeling

MrShark:
Adding labels to parts, starting from the top and going down, skipping things that I'm not sure about the labeling

### DIFF
--- a/boxes/generators/angledbox.py
+++ b/boxes/generators/angledbox.py
@@ -35,7 +35,7 @@ class AngledBox(Boxes):
             choices=["none", "angled hole", "angled lid", "angled lid2"],
             help="style of the top and lid")
 
-    def floor(self, x, y, n, edge='e', hole=None, move=None, callback=None):
+    def floor(self, x, y, n, edge='e', hole=None, move=None, callback=None, label=""):
         r, h, side  = self.regularPolygon(2*n+2, h=y/2.0)
         t = self.thickness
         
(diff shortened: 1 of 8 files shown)


boxes/generators/angledbox.py
```diff
@@ -108,36 +108,41 @@ def render(self):
...
         if i % 2:
             self.rectangularWall(lx, h, move="right",
-                                 edges=b+"gfg" if fingers else b+"geg")
+                                 edges=b+"gfg" if fingers else b+"geg",
+                                 label="wall {}".format((n+1)*2))
         else:
             self.rectangularWall(lx, h, move="right",
-                                 edges=b+"GfG" if fingers else b+"GeG")
+                                 edges=b+"GfG" if fingers else b+"GeG",
+                                 label="wall {}".format((n+1)*2))
```
florianfesti, line 145: If the front wall is 1 shouldn't the back wall be n+2  with the second set of walls having larger numbers. May be we should even squeeze the back wall in between the right and the left end pieces...

MrShark: Of course it should. Changed the numbering, but did not rearrange the walls (could not think of a way to do it without duplicating the inner loop or adding ugly branches that makes the logic hard to follow)

florianfesti:
This is the way 2154245cde0a88505d240620d34577d2e6594366
It's even shorter. Although is has some weird logic.

Anyway, thanks for all the work.

PR #335, opened 2021-11-03
fix some typos with codespell

matkoniecz:
(codespell found more but then I realized that I am procrastinating, so submitted only what I fixed already)


florianfesti:
Thanks for the fixes (so far)!

Issue #337, opened 2021-11-05
Question: interested in code formatting?


javl:
Before I start making any changes I was wondering if you'd be interested in a reformatted version of boxes.py, using something like the pycodestyle linter. No change in functionality, just an update to use f-strings and some more consistent formatting.
Also, a bunch of files currently don't have a file extension (especially in the scripts folder). While this is valid on Linux, it is kind of odd and it would be nice to add them.

florianfesti:
Although the project is Python 3 nowadays most of the code base very much is still Python 2. Also a lot of pieces were done rather quickly. So reformatting and converting to a more modern style is welcome. This should obviously be done in separate commits that don't do functional changes.
File extensions are a bit of a different topic. They are frowned upon for executables on Unix systems. So I'like to keep the executable names there. But I am open to install them with extension on Windows or may be even rename them but add symlinks on Unix.

javl:
Do you have any references for this last comment, on file extensions being frowned upon? I've never seen this in any other project (been running Linux for years).

florianfesti:
OK; may be frowned upon is a bit too harsh. A look into /usr/bin shows there actually are a few executables that use file extensions. But most don't - even when only looking at scripts. Guess this has become more acceptable in recent years -  with Python being the worst offender. I generally think a file extension leaks an implementation detail into the (CLI) interface. Which is something to be avoided.

Otoh as the name of the project is `Boxes.py` I should probably reconsider my POV...

florianfesti:
Guess the answer is "Not as much as I first thought"

PR #343, opened 2021-11-11
Update formatting of all code

javl:
As mentioned in #337 I'd like to update the codebase to use more consistent formatting and address some other quality of life updates. I feel this will make it a lot easier to expand this (already great) project.
I pulled the latest version of master just before this PR (thats why there are two commits) so you should be able to merge without any problems if you do decide to merge.

To be able to maintain some overview while making changes to this many files this PR contains ONLY the following changes:
* Remove unneeded white space in all files (like at the end of lines, or multiple newlines at the end of files)
* Reformats all python code using the Pycodestyle linter. 
* Adds a config file (setup.cfg) for the linter. 
 
 
You can automatically apply the linter using autopep8 on the commandline (`autopep8 --in-place`) from the main directory. If you're using VS Code to edit code, you can enable the Pycodestyle linter and add this to the config in .vscode/settings.json to use the same setup:
```
{
  "python.linting.pycodestyleEnabled": true,
  "python.linting.pycodestyleArgs": [
    "--max-line-length=180",
    "--ignore=W503,E741"
  ],
  "python.linting.enabled": true
}
```

The two Pycodestyle warnings/errors I opted to ignore are W503 (no arithmic operators (+-/ etc.) at the beginning of lines) and E741 (no variables named l o i (as `l` is used in some places)
 
Some changes I'd like to make in the future (but NOT in this commit):

* move to f-strings to improve readability
* remove old comments
* Add file extensions to all files


javl:
As a note: just realized I didn't update the docs, but you might want to make changes before doing so anyway.

static/self.css
```diff
@@ -6,7 +6,7 @@ button,
 input,
 select,
 textarea {
-    color: #222;
+    color:  #222;
```
Rotzbua, line 9: `*.css` should be ignored

javl: Not sure where this space came from. Autopep8 will by default only run on .py files. Think I might have messed up while merging some edits made in between my update and creating the PR. Can't really prevent having at least one error like this when you edit 100+ files ;)

florianfesti:
OK; I tried to rebase and review the change but this is not that easy. Can you split the automatically generated changes into a separate patch so I can just run `autopep8` myself and see it creates the same changes, please? Also there are changes like updating the language files in there that - if at all - need to go into a separate patch.
Either you are using some other version/setting for `autopep8` than me or there are some manual changes. If they are they also need to go into a separate patch.

Sorry, for the late reply!

florianfesti:
OK, I have played around with this a few times but never was really happy enough to merge it. For one the reformatting generates a lot of lines longer than 80 characters. The change (beside being very outdated now) is basically impossible to review properly. I guess I am just giving up and that for now. Closing.

javl:
@florianfesti Personally I'd let go of the 80 char width, as I don't really think its needed on modern systems (I've been using 130 for a while now, not sure where I picked up that number). But I guess that's also a matter of preference. But yeah, with so many changes it's hard to review and make sure nothing broke :/

Issue #351, opened 2021-12-17
A PaperBox that needs less paper


fuddl:
I tried to recreate a chocolate box with the dimensions 85×84×13 mm. Unfolded, this  box could be made out of a sheet of 133×199 mm.

![an unfolded paper box](https://user-images.githubusercontent.com/842548/146583615-21803321-4629-4379-bc9c-cb6a4cc7f5fc.png)

If I generate a PaperBox [with the same dimensions](https://festi.info/boxes.py/PaperBox?x=84&y=13&h=85&lid_heigth=15&lid_radius=7&lid_sides=20&margin=0.5&mark_length=1.5&tab_angle_rad=0.07982998571223732&finger_hole_diameter=15&thickness=3.0&format=svg&tabs=0.0&debug=0&labels=0&labels=1&reference=100&burn=0.1&render=1) it needs sheet of 213×294 mm with a lot of paper to throw away after. Also if scaled up, it exceeds the maximum paper size of my plotter. 😅

I propose a more space efficient PaperBox design based on my chocolate box.

Box                   | sheet size needed
---------------|---------------------
PaperBox         | 62 622mm²
[new design]  | 26 467mm²

The new design requires 42.26% less paper. Is it a good idea? Is that something worth looking into?

gcollic:
Hello !

I'm the original contributor of this paper box. Here is my opinion ;-)

Is the original paper box like you want if you exchange y and h ? https://festi.info/boxes.py/PaperBox?x=84&y=85&h=13&lid_heigth=15&lid_radius=7&lid_sides=20&margin=0.5&mark_length=1.5&tab_angle_rad=0.07982998571223732&finger_hole_diameter=15&thickness=3.0&format=svg&tabs=0.0&debug=0&labels=0&labels=1&reference=100&burn=0.1&render=1

Your box is similar but doesn't look identical.The current box was created for advent calendar cases, it's not clear if I can use your design for this purpose without some tradeoff, and I wonder if your design is more efficient with any parameters combination (in paper size, but also folding simplicity, and opening ergonomy).

I think having both could maybe be interesting. I'm not sure if it would be better as a whole new box or as a simple "design type" parameter for the current box. In both cases, we could think collectively of how to name both so that the difference makes sense ;-)

gcollic:
[...]

PR #352, opened 2021-12-25
Paperbox: second design suggested in #351

gcollic:
Second design suggested in #351 for the Paperbox.

@fuddl here is the result for `--x=84 --y=13 --h=85 --design=tuckbox` [paperbox.zip](https://github.com/florianfesti/boxes/files/7776026/paperbox.zip)
Do you have suggestions? For example on the design parameter name, descriptions, or on explanatory sketches?

### DIFF
--- a/boxes/generators/paperbox.py
+++ b/boxes/generators/paperbox.py
@@ -16,7 +16,6 @@
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import math
-from functools import partial
 from boxes import Boxes
 
 
--- a/boxes/generators/paperbox.py
+++ b/boxes/generators/paperbox.py
@@ -43,6 +43,15 @@ class PaperBox(Boxes):
         self.buildArgParser("x", "y", "h")
 
         self.argparser.add_argument(
+            "--design",
+            action="store",
+            type=str,
+            default="automatic",
+            choices=("automatic", "widebox", "tuckbox"),
+            help="different design for paper consumption optimization. The tuckbox also has locking cut for its lid.",
+        )
+
+        self.argparser.add_argument(
             "--lid_heigth",
             type=float,
             default=15,
@@ -64,7 +73,7 @@ class PaperBox(Boxes):
             "--margin",
             type=float,
             default=0.5,
-            help="Space between the two folded sides to glue",
+            help="Margin for the glued sides",
         )
         self.argparser.add_argument(
             "--mark_length",
@@ -87,9 +96,63 @@ class PaperBox(Boxes):
         )
 
     def render(self):
-        width = self.x
-        length = self.y
-        height = self.h
+        if self.design == "automatic":
+            self.design = "tuckbox" if self.h > self.y else "widebox"
+
+        path = (
+            self.tuckbox(self.x, self.y, self.h)
+            if self.design == "tuckbox"
+            else self.widebox(self.x, self.y, self.h)
+        )
+
+        self.polyline(*path)
+
+    def tuckbox(self, width, length, height):
+        lid_cut_length = min(10, length / 2, width / 5)
+        half_side = (
+            self.mark(self.mark_length)
+            + [
+                0,
+                90,
+            ]
+            + self.ear_description(length, lid_cut_length)
+            + [
+                0,
+                -90,
+                length,
+                0,
+            ]
+            + self.lid_cut(lid_cut_length)
+            + self.lid(width - 2 * self.thickness)
+            + [0]
+            + self.lid_cut(lid_cut_length, reverse=True)
+            + [
+                length,
+                -90,
+            ]
+            + self.ear_description(length, lid_cut_length, reverse=True)
+            + self.mark(self.mark_length)
+        )
+        return (
+            [height, 0]
+            + half_side
+            + self.side_with_finger_hole(width, self.finger_hole_diameter)
+            + self.mark(self.mark_length)
+            + [
+                0,
+                90,
+            ]
+            + self.tab_description(length - self.margin - self.thickness, height)
+            + [
+                0,
+                90,
+            ]
+            + self.mark(self.mark_length)
+            + [width]
+            + list(reversed(half_side))
+        )
+
+    def widebox(self, width, length, height):
         half_side = (
             self.mark(self.mark_length)
             + [
@@ -124,18 +187,22 @@ class PaperBox(Boxes):
                 0,
             ]
             + self.mark(self.mark_length)
-            + [
-                self.lid_heigth - self.lid_radius,
-                (90, self.lid_radius),
-            ]
         )
-        path = (
+        return (
             self.side_with_finger_hole(width, self.finger_hole_diameter)
             + half_side
-            + [width - 2 * self.lid_radius]
+            + self.lid(width)
             + list(reversed(half_side))
         )
-        self.polyline(*path)
+
+    def lid(self, width):
+        return [
+            self.lid_heigth - self.lid_radius,
+            (90, self.lid_radius),
+            width - 2 * self.lid_radius,
+            (90, self.lid_radius),
+            self.lid_heigth - self.lid_radius,
+        ]
 
     def mark(self, length):
         if length == 0:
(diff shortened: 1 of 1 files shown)


fuddl:
@gcollic that's super interesting. I managed to set up your branch and here is my first prototype. it houses a computer mouse.

`./scripts/boxes paperbox --x=63 --y=38 --h=116 --design=tuckbox`
a paper box that houses a computer mouse

it is made from cheap former food packaging. The closed side is quite stable and it felt like it didn't even need glue to keep its shape. But the weak part is here:

a worn out edge of a paper box

I assume the `thickness` parameter is ignored in PaperBox. I think it might be helpful here:

The _ears_ need a little gap to allow the lid to come in: otherwise they quickly wear out 



What do you think?

gcollic:
You are right, I should look into that, and try to change the default
thickness for this box especially if feasible

fuddl:
second try. I have a question. What is the point of these cuts?



Again, this is where a lot of stress happens:


I would replace them by a single debossed line.

gcollic:
They are from your chocolate box, it helps prevent accidental opening of the box.
image

I thought you wanted them, but if not I can simply remove them :-) I initialized them at half the length of the box, but considering what it does given these parameters, it was obviously a bad choice ^^

Your cuts are not burnt, I'm wondering what do you use ? A cricut maybe ?

fuddl:
> I thought you wanted them, but if not I can simply remove them :-) 

I don't know what I want. I'm trying to figure out why they didn't work in our case 😅

> They are from your chocolate box, it helps prevent accidental opening of the box.

sound good. Someone suggested to me that the reason, the slids exist is to determine if the box has been opened before, because it tends to look messy after opening.

I made an excursion to muji where they sell differently sized products in identically designed boxes and the slids always seem to have the same length: not more then a cm. So the length of the slids should probably max out somewhere. maybe depending on the material 🤷



But I'll try with and without slid and we'll see, what the outcome is.

> Your cuts are not burnt, I'm wondering what do you use ? A cricut maybe ?

Yes, I'm using a cricut maker. 😉 How could you tell? How do you plot paper boxes?

boxes/generators/paperbox.py
```diff
@@ -177,3 +254,25 @@ def tab_description(self, height, width):
             side,
             deg - 90,
         ]
+
+    def dented_tab_description(self, width, reverse=False):
+        deg = math.degrees(self.tab_angle_rad)
+        side = width / math.cos(self.tab_angle_rad)
+        end_width = width - 2 * width * math.tan(self.tab_angle_rad)
```
fuddl, line 261: these three lines can be removed

boxes/generators/paperbox.py
```diff
@@ -177,3 +254,25 @@ def tab_description(self, height, width):
...
+            0,
+            (-90, width / 4),
+            width / 4,
+            90,
+            width - (width / 4) - 4 * self.burn,
+            90,
+            width,
+            -90,
+            2 * self.burn,
+        ]
```
fuddl, line 276: please try this instead:
```py
        path = [
            2 * self.burn,
            -90,
            self.lid_heigth - self.lid_radius,
            (90, self.lid_radius),
            width - self.lid_heigth - 4 * self.burn,
            90,
            self.lid_heigth,
            -90,
            2 * self.burn,
        ]
```
I tried to give the _ears_ the same length as the lid

boxes/generators/paperbox.py
```diff
@@ -87,9 +96,61 @@ def __init__(self):
...
+                90,
+            ]
+            + self.dented_tab_description(length)
+            + [
+                0,
+                -90,
+                length,
+                0,
+            ]
+            + self.lid_cut(length / 2)
```
fuddl, line 124: 1 cm maximum
```py
+ self.lid_cut(min(length / 2, 10))
```

boxes/generators/paperbox.py
```diff
@@ -87,9 +96,61 @@ def __init__(self):
...
+            + self.lid(width)
+            + [0]
+            + self.lid_cut(length / 2)
```
fuddl, line 127: 1 cm maximum
```py
+ self.lid_cut(min(length / 2, 10))
```

fuddl (commented):
I tried to address some of the changes I proposed. The lid cuts are limited to 1cm and the ears have the same dimensions as the lid

gcollic:
> Yes, I'm using a cricut maker. 😉 How could you tell?

Well, I just guessed because it's the only hobbyist paper plotter I know ^^

> How do you plot paper boxes?

Since boxes.py is a laser cutter template generator, I do use a laser cutter ^^ An emblaser 2 :-) It's not the best tool to cut paper since it burns it, but it's what I have, and definitely better than cutting by hand in some cases. When my wife planned only 2 weeks in advance to create 5 advent calendars and all their contents for the family, it also meant 120 precise paper boxes, and it would not have been possible to finish all in time without this (we just needed to clean the edge before folding), and even if it would, it never would have been as precisely closing and professionally looking if made by hand ;-) A paper plotter would definitely have been better, but I don't cut paper often enough to invest in one ;-)

boxes/generators/paperbox.py
gcollic: Good point, thanks!

boxes/generators/paperbox.py
gcollic: The point of the cut in the lid was related to the previous shape, I feel like now it can't lock sharply like it's purpose was: there is no link between the cut length and the ear, and the lid is sliding against a soft curve in the ear instead of a right angle locking it.

I would personnaly suggest to either remove the cut and keep the soft angle, or

1. use a variable to define the lid_cut_size (length/2, but maximum 1cm, and maximum a fifth of the lid) and use it both in the lid cut and the ear design
2. change a bit this design so that the cut and the ears would lock according to this common information with a right angle

What do you think ? I will push something like this to illustrate

gcollic: I also incorporated the paper thickness in the new design :-)

fuddl:
I tried the mouse box again:
`./scripts/boxes paperbox --x=63 --y=38 --h=116 --design=tuckbox --thickness=.5 --mark_length=0 --reference=0`

We are getting somewhere. It feels really solid! 🥳

I just had to make a tiny modification: the thickness is not taken into account for the lid's _tongue_:

IMG_0910

After removing .5mm from each side it fits perfectly:

IMG_0911

Also I like that you renamed `dented_tab` to `ear` 😅

gcollic:
Thank you for your tests ! Much appreciated :-)

fuddl:
@gcollic look at this beauty (same parameters as in the last iteration)

IMG_0916

just one thing which might make assembly easier: the one thing the human has to do _very precisely_ is to place and glue the tab at exactly one paperwidth away from the back wall:

hard

It might be easier if the human only had to align the tab with the back wall:

easier

Does it make sense? But this is a minor improvement. Feel free to merge. Thanks for your patience. 😃

gcollic:
There is a margin parameter in the previous box, I tried to make it works
logically with this one. Do you have what you want with a margin of 0 ? It
could be the default value for both boxes

fuddl:
> There is a margin parameter in the previous box, I tried to make it works logically with this one. Do you have what you want with a margin of 0 ?

I tried it with this box for an external hard drive:
`./scripts/boxes paperbox --x=90 --y=30 --h=136 --design=tuckbox --thickness=.5 --mark_length=0 --reference=0 --margin=0`
external drive box

>  It could be the default value for both boxes

I would recommend that.

If you ask me, this feature is complete 👍

I hope you have a use for what you made here.

gcollic:
@florianfesti we have finished modifying the code, we are ready for you to merge and/or doing feedbacks :-)

Thank you !

florianfesti:
I have watched the progress of this PR with great joy. That's a lot for your work and dedication!
I adjusted the commit message a bit so it does not rely completely on the GH ticket and attributes fuddl's contribution. I hope that's fine with you.

I'll update the online instance soon. Still have a few other things I want to look into first.

gcollic:
> I hope that's fine with you.

Of course! :-)

I forgot, is it possible to change the default value of common shared parameters such only for a specific box, such as the "thickness" parameter only for the paperbox ?

fuddl:
If you need reference photos or explanatory renderings, let me know

florianfesti:
Parameters are just an parseargs object (`self.argparser``). So it should be possible to change that in the constructor of the generator. `.set_defaults()` is one way to achieve this. You may also be able to fetch the argument object and modify this. But I have not dug into the argparse docs that far.

May be an update to the description and a few more or new pictures are also helpful. The ones in https://github.com/florianfesti/boxes/pull/352#issuecomment-1003372862 or https://github.com/florianfesti/boxes/pull/352#issuecomment-1003559235 ready look pretty nice

florianfesti:
I already fixed the formatting of the description which was in a fixed font due being indented.

PR #361, opened 2022-01-20
Update slatwalldrillbox.py

zjzh:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


florianfesti:
OK, I am a bit old school but I just don't think this is an improvement. Yes, it does use more Python features but for me at least it is much harder to read. The `:1` at the end of the slice ranges is even unnecessary. But all in all this is just less readable.

I am all for better code and better code style and there are a lot of places in Boxes.py that clearly show outdated and not that great Python2sms. But this is not the place.

PR #363, opened 2022-02-04
remove Flex Settings

HaSHsss:
This parameter is not used in any box


florianfesti:
The flex settings are used by the chest lid. But once could argue that it doesn't really belong here. But as long as it is the flex settings are needed.

HaSHsss:
Thanks for the clarification, it's clear now
Is it possible move LID settings (additional lid (for straight top_edge only)) to the top_edge  list?

florianfesti:
Unfortunately this is not that easy `x` down to `outside` are standard options that are all added at once. `vertical_edges` and `lid` can both only added after them ( https://github.com/florianfesti/boxes/blob/master/boxes/generators/universalbox.py#L33). In theory all those settings cold be added by hand, but I probably can't the bothered. The whole lid situation needs a better solution in the long run anyway.

PR #375, opened 2022-03-27
add Lightburn .lbrn2 file format support

SuksAE:
This adds native file output support for Lightburn (https://lightburnsoftware.com/) file format. The code is used for creating my projects since about 3 months now, so the most obvious errors have been fixed.

Unfortunately there is no official documentation for the file format of Lightburn, so the implementation relies on reverse engineering of saved files. Furthermore this is not officially supported by Lightburnsoftware (they don't even know this exists).

Thanks
  Klaus Steinhammer

PS: as I am new to Github and this is the first time submitting some code this way, please be patient with me.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1279,7 +1279,7 @@ class Boxes:
             self.corner(-a)
 
     @restore
-    def text(self, text, x=0, y=0, angle=0, align="", fontsize=10, color=[0.0, 0.0, 0.0]):
+    def text(self, text, x=0, y=0, angle=0, align="", fontsize=10, color=[0.0, 0.0, 0.0], font="Arial"):
         """
         Draw text
 
@@ -1315,7 +1315,7 @@ class Boxes:
                 raise ValueError("Unknown alignment: %s" % align)
 
         for line in reversed(text):
-            self.ctx.show_text(line, fs=fontsize, align=halign, rgb=color)
+            self.ctx.show_text(line, fs=fontsize, align=halign, rgb=color, font=font)
             self.moveTo(0, 1.4 * fontsize)
 
     tx_sizes = {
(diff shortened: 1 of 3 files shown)


florianfesti:
This looks really nice! Guess we finally have to update our Lightburn instance...

There are a few minor nitpicks that can be fixed later on. Not 100% sure if I like the way it uses the coordinate system. Otoh there is an argument to be made that showing the drawing upside down actually is the righ thing to do if the coordinate system of the laser is aligned that way.

On Linux I currently get an error that Arial is not found. The whole text system needs some love no matter what. So this is only a minor annoyance.

These do not detract from the fact that this is a really cool feature and the patch looks good overall. Thanks a lot!

SuksAE:
> This looks really nice! Guess we finally have to update our Lightburn instance...
> 
> There are a few minor nitpicks that can be fixed later on. Not 100% sure if I like the way it uses the coordinate system. Otoh there is an argument to be made that showing the drawing upside down actually is the righ thing to do if the coordinate system of the laser is aligned that way.
> 
> On Linux I currently get an error that Arial is not found. The whole text system needs some love no matter what. So this is only a minor annoyance.
> 
> These do not detract from the fact that this is a really cool feature and the patch looks good overall. Thanks a lot!

In the beginning I was also a little annoyed about the coordinate missmatch with Lightburn. But as I later found out, it depends on the setup of the origin from your machine. If the origin is located on the front left, then the orientation of lbrn2 output is exactly the same as with the imported svg file. If the origin is located back right, the both X and Y coordinates are mirrored.

But I thin that in the end the parts are cut from sheet material, where most of the time it is not important which side is front and which is back on the finished piece.

If it is important for me, then I either 
- have to mirror the generated files according to my machine setup, or 
- knew if i have to place the front side of the material up or down to get the orientation I need.

Creating an automatism for mirroring would require knowledge about the machine the file is intended for, and this is way too much effort...

I wonder how this is handled by people owning multiple laser machines with different origin setups? I guess they can't use the same file on all of their machines without adjustment...

regards
  Klaus

PR #376, opened 2022-03-27
Initial commit of Bottle Tag generator

smvoigt:
I'm not sure you want this sort of paper base project in here but I found that this provided a good base to create the bottle tag that I wanted. I won't be offended if you don't want it.

This is a simple tag that can go over different size bottles. You can print information on the paper prior to cutting.


florianfesti:
Well, this is another paper generator already - PaperBox. While I am personally not that interested in paper as a material and the core functionality is for focused around sheets there is no reason to reject generators like this. Two or three more and they can get their own group and move out of "Misc".

`r1` and `r2` could probably have better names and probably should denote the diameter which is more natural for round objects. Yes, I am not the one that should point fingers wrt parameter names, just pointing it out....

florianfesti:
Adjusted the parameters in 94935d96bd5036b28bfdff78048ad9c83c065f38. Thanks for the patch!

smvoigt:
Thank you for doing that. I was going to address it over the weekend - I'm happy with what you have done.

PR #377, opened 2022-03-27
Feat/divider tray outside walls finger notches

gcollic:
Nearly every time I'm using the DividerTray I contributed, I end up adding the same notches in the walls than the dividers, to help catch the content, so here is this modification.

From my experience it should be the default, but I can live with it being opt-in ;-)

I'm not really fond of this new edge object not using the actual length received in parameter. I could adjust the "sx" in case there is a mismatch, but there is no call to this edge needing to add this code for now. Do you have an opinion on this?

Thank you!

### DIFF
--- a/boxes/generators/dividertray.py
+++ b/boxes/generators/dividertray.py
@@ -32,6 +32,12 @@ class DividerTray(Boxes):
         self.addSettingsArgs(edges.FingerJointSettings)
         self.buildArgParser("sx", "sy", "h", "outside")
         self.argparser.add_argument(
+            "--notches_in_wall",
+            type=boolarg,
+            default=True,
+            help="generate the same notches on the walls that are on the dividers",
+        )
+        self.argparser.add_argument(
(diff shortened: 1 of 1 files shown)


florianfesti:
Well, such trickery is not completely unheard of for special Edge classes like this. Yes, its ugly.  No, I don't care enough to come up with a better solution.

Anyway, this is a pretty nice add-on.

I wonder if the settings for the slots and diviers really should go into a separate Settings object. An be it only to clean up the UI. But this is beyond the scope of this patch - even if it makes the problem of cluttered UI a tiny bit worse.

Thanks for your work!

gcollic:
You are totally right! Would you want the settings in `dividertray.py` or `edges.py`?

florianfesti:
I would put them in dividertray for now. Turns out I have used up most of the letters of the alphabet so adding new edges is a bit of an issue. Need to clean that up at some point and move things like the hinges (that are made from 4 or 6 edge types) out of the global name space.
We can move them to edges.py later on if there are more generators that would like to use them.

PR #378, opened 2022-03-27
feat: DividerTray make the play optional

gcollic:
The new features on the DividerTray are great, thank you ^^

There is now a play for easier divider insertion and removal, but in my case the lack of play was intentional and can help structural integrity with longer box insert made from more flexible wood (the dividers are rarely moved), so I propose to make the play optionnal.


florianfesti:
Yeah, I guess it depends a bit on the use case or whether you want to configure the tray once or whether you have changing amount of stuff and need to adjust the dividers all the time.

PR #379, opened 2022-03-28
Lbrn2 output

SuksAE:
This escallated quickly - got the first feedback after 5 hours.

User is confused about export of empty text fields. 

There are some generators, which do not set labels for their parts. Unfortunately per default a label is set to an empty string, which when output this way creates an empty text field. In an SVG file this also happens, but is not so outstanding than in Lightburn.

So I added a check if a text field is empty, in which case the output of the field is skipped...

Please add to main branch.

Thanks
 Klaus

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1279,7 +1279,7 @@ class Boxes:
             self.corner(-a)
 
     @restore
-    def text(self, text, x=0, y=0, angle=0, align="", fontsize=10, color=[0.0, 0.0, 0.0]):
+    def text(self, text, x=0, y=0, angle=0, align="", fontsize=10, color=[0.0, 0.0, 0.0], font="Arial"):
         """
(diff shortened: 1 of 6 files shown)


florianfesti:
Merged as cf99c93400b88c4263a2db253e4727740db2eb5b

florianfesti:
For smaller changes like this it is preferred to not use `git merge` but to `rebase` changes onto the current `master` branch. This prevents those ugly merge commits and makes it easier to spot issues created from parallel branches.

But there are many ways to use git but I found this to work best for a pull request based workflow.

florianfesti:
I added 70c49afb57b312b9d1fa35e1cd51b75b3f539ff1 to prevent the issue for other formats. Technically the defaults for labels should be `None` to have a clean way to represent no label - but I could not be bothered...

PR #380, opened 2022-03-28
Feat/divider tray settings

gcollic:
See #377 
> I wonder if the settings for the slots and diviers really should go into a separate Settings object. An be it only to clean up the UI.

Here is a proposition of settings. Divider play is now relative to thickness again 😃


florianfesti:
I was about to nitpick a few things. But I just hit merge and changed a few descriptions in 84797abc356a26d97db3974569e76fe362ab3b63 instead.

PR #381, opened 2022-03-30
fixed error generating an empty path list

SuksAE:
Found an error which broke lbrn2 output.

In function _faster_edges(self, inner_corners)_ the duplicate removal fails if there is only one element in the path list. This single element gets removed in any case, which results in an empty path list, which in turn will crash the lbrn2 output.

Furthermore trying to output an empty list via print also results in an error which costs me some time at debugging...

This patch deals with both issues.

Cheers
  Klaus

### DIFF
--- a/boxes/drawing.py
+++ b/boxes/drawing.py
@@ -150,8 +150,10 @@ class Path:
     def __repr__(self):
         l = len(self.path)
         # x1,y1 = self.path[0][1:3]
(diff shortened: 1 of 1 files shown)


florianfesti:
Yeah, the back end still needs some smoothing. We probably should filter out empty paths and path with only one entry (that is a move command). Being defensive about that is a good thing though. Thanks for the patch!

PR #386, opened 2022-04-23
Mounting edge

SuksAE:
I needed to mount several box design to the wall. Therefore I had to add mounting holes in post processing. As this was cumbersome, I created a mounting hole drawing function and an appropriate edge type. I also updated some box generators to use these functions.

I hope this is also usefull to others.

Cheers
  Klaus


florianfesti:
OK, there is a lot going on here. These changes need to be split into meaningful commits. E.g. all the work on labels needs to be separate. One could argue the this should be 3 separate pull requests but let's not be too petty.

I am not a big fan of the MountingEdge. The mounting holes are an orthogonal feature to the edge being used and can be useful on any generator independently of what edge types are being used. So using callbacks everywhere is they way to go. You can create a callable object or add callback methods to the Boxes class to make this easier and available everywhere without copying code. Settings can also be used without edges attached to them.

The text placement issue is a bit more complicated than you probably anticipated. The problem here is that "middle" is not properly supported everywhere - especially on Firefox - while working on Chrom(ium). This needs to be fixed by using a different alignment. You just move the breakage to another browser.

lids.py should probably not just pass through the label parameters but actually apply its own. May be allow passing a prefix to the label but I doubt that's necessary right now.

Don't let this dishearten you. There is a lot of great changes in this PR.

florianfesti:
If you have trouble with splitting the commits in git I am happy to help or do the work for you. It's a bit of an advanced skill and a bit scary for new comers when undoing the commit with `git reset` for the first time.

SuksAE:
Yes, please. Help is always apprechiated.

What do you suggest how we can get this to the main branch?

florianfesti:
Just start with the necessary changes to the code. We can get the commit into shape afterwards. In general having smaller commits makes things easier as moving them around and combining them is pretty easy with `git rebuild -i`.

Splitting a commit can be done like this: https://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/
For beginner I'd suggest to duplicate the branch first so you can have a look at the version before the change for comparison.
Another very useful command is `git add -p` that allows you to add single lines to a commit while leaving other changes in the file to be committed later.

Massaging commits is a valuable skill to learn IMHO. But if you don't feel comfortable with that I can also do it.

florianfesti:
Looks like this is super seeded by your other two PRs. May be just close this one.

SuksAE:
Before we close this PR, how shall we continue with the MountingEdge feature?

I like it because it implements mounting holes to many different generators without the need of modifying each one separately.
If I would implement the mounting holes to each generator which could potentially use the MountingEdge, it would result in a lot of copying and pasting of identically code to several files...

What is your idea regarding this topic?

PR #389, opened 2022-04-30
Add host parameter to boxesserver

anisimov:
Ability to specify host for running web server. 
e.g. ./boxesserver localhost:8000

This userful for start under nginx. For start boxesserver in VDS with several virtualhost.


florianfesti:
Thanks for the patch!

PR #392, opened 2022-05-13
Added pear shaped mounting hole drawing function

SuksAE:
Separated the mounting hole drawing functions from the previous pull request.


boxes/__init__.py
```diff
@@ -1278,6 +1301,38 @@ def flatHole(self, x, y, r=None, d=None, w=None, rel_w=0.75, angle=0):
...
+        
+        if d_shaft < 2*self.burn:
+            d_shaft = 2 * self.burn + 1E-9
+        rs = d_shaft/2
+        rh = d_head/2
+        
+        if not d_head or d_head < 2* self.burn:
+            self.hole(x,y,d=d_shaft,tabs=tabs)
+            return
+
```
florianfesti, line 1326: This makes the hole rotate around the top point instead of the center when modifying the angle. Use two .moveTo calls. The first to move to the position of the hole center pointing in the right direction and then move out to the start of the line. That way the radius moving to the line points in the right direction.

boxes/generators/trafficlight.py
```diff
@@ -62,6 +62,9 @@ def __init__(self):
         self.argparser.add_argument(
             "--upright",  action="store", type=boolarg, default=True,
             help="stack lights upright (or side by side)")
+
```
florianfesti, line 65: I wonder if we can get away with only one argument and don't add holes if it is left empty or one of the diameter(s?) are zero. This is how we handle it with several other features already. Having the holes enabled by default is fine IMHO. For one it brings attention to this features to the users. It is also kinda necessary to have a good example on how to use it.

Given the format is more complex than just a number one could add a section in the docs and add an Info link as for some other parameters. But that's not strictly necessary.

boxes/generators/trafficlight.py
```diff
@@ -85,13 +88,13 @@ def topCB(self):
     def frontCB(self):
         self.hole(self.h/2, self.h/2, self.h/2-self.thickness)
 
-    def wall(self, h1, h2, w, edges="ffef", callback=None, move=""):
```
florianfesti, line 88: Coding style nitpick: No white space around the equal sign for named parameters (in opposite to normal assignment statements).

boxes/generators/trafficlight.py
```diff
@@ -126,44 +165,46 @@ def render(self):
...
-        self.rectangularWall(th, h, "FFFF", callback=[self.backCB], move="up")
+        if self.upright:
+            self.rectangularWall(th, h, "FFFF", callback=[self.backCB, self.addMountV(th,h)], move="up", label = "back")
+        else:
+            self.rectangularWall(th, h, "FFFF", callback=[self.backCB, self.addMountH(th,h)], move="up", label = "back")
 
         if self.upright:
             # sides
-            self.rectangularWall(th, d, "fFsF", callback=[self.sideCB], move="up")
-            self.rectangularWall(th, d, "fFsF", callback=[self.sideCB], move="up")
```
florianfesti, line 134: Generally these new labels that are unrelated to the new mounting holes should be in a separate commit (and may be even pull request). MAy be I do not quite care enough to split them out, though.

build_number
```diff
@@ -0,0 +1 @@
+1308
```
florianfesti, line 1: Are you adding these numbers or is this a new GitHub thing?

boxes/__init__.py
```diff
@@ -1237,6 +1237,17 @@ def rectangularHole(self, x, y, dx, dy, r=0, center_x=True, center_y=True):
     @restore
     @holeCol
     def dHole(self, x, y, r=None, d=None, w=None, rel_w=0.75, angle=0):
```
florianfesti, line 1239: This doc strings technically belong in their own commit. No big deal though.

florianfesti (changes requested):

boxes/generators/trafficlight.py
florianfesti: Oh, I did not quite realize that this is a re-work of your first PR. Didn't intend to continue whiling about this. Sorry, no second coffee yet.

boxes/__init__.py
florianfesti: Same as above.

build_number
SuksAE: Its kind of both - you can configure git to create this automatically. I am looking for a way to get a version string to be shown at the html page, but for now it is kind of half baken. This got accidentally to this pull request - i disabled the feature again and removed the file - ignore it for now. Maybe I come back to you for this feature at a later point in time...

boxes/__init__.py
```diff
@@ -1324,7 +1324,8 @@ def mountingHole(self, x, y, d_shaft=0.0, d_head=0.0, angle=0, tabs=0):
             self.hole(x,y,d=d_shaft,tabs=tabs)
             return
 
-        self.moveTo(x + rs - self.burn, y, angle)
+        self.moveTo(x,y,angle)
```
florianfesti, line 1327: `rs - self.burn` moved from x to y direction. You sure about that? Also one space after commas, please!

boxes/__init__.py
```diff
@@ -1278,6 +1301,39 @@ def flatHole(self, x, y, r=None, d=None, w=None, rel_w=0.75, angle=0):
...
+
+        :param x: position
+        :param y: postion
+        :param d_shaft: diameter of the screw shaft
+        :param d_head: diameter of the screw head
+        :param angle: rotation angle of the hole
+
+        """
+        
+        if d_shaft < 2*self.burn:
```
florianfesti, line 1318: May be the check for the  shaft being zero should go here instead of being duplicated in every generator.

boxes/generators/trafficlight.py
florianfesti: Still one space to much in line 96

florianfesti (changes requested):
Looks pretty good already. After these minor nitpick this should be ready to go in.

I am wondering whether it's worth adding the hole_dD parameter to .buildArgParser() but that's something we can decide on later.

boxes/__init__.py
SuksAE: yesss, thats my one of my superpowers - unintentionally swap x and y coordinates ;-)

now "angle" rotates arround the center of the shaft hole...

florianfesti:
Thanks for all the work to clean this up! I hope my nitpicking is not too annoying

SuksAE:
> Thanks for all the work to clean this up! I hope my nitpicking is not too annoying

Nah - I understand that a consistent code base is a necessary evil if different people shall work on a project...

PR #396, opened 2022-05-26
Changed svg text alignment

SuksAE:
According to https://stackoverflow.com/a/21373135 this should fix text alingment in Firefox without negative influence for others.


HaSHsss:
I didn't know about this bug.
But your solution works great.

florianfesti:
Thanks a lot. I should have looked into that a long time ago. Merged with slightly modified commit message. Important pointers like the Stackoverflow link should go into the commit message so it can be found via git blame. The messages in the PRs are not linked to the code in any meaningful way unfortunately.

Issue #330, opened 2021-09-13
Request: Can dispenser generator


AshaiRey:
I would like to request a new generator. One to generate a can dispenser at various sizes.
The size would depend on the can size and the number of cans it would hold in 2 layers.
And example of a can dispenser can be seen here 
https://d2t1xqejof9utc.cloudfront.net/screenshots/pics/04c43617455bef35938629dcf7d501da/large.JPG

florianfesti:
43bd7d12d318e353cc1178f714701a23ef6ed989 implements this. This is the result of #401

https://festi.info/boxes.py/CanStorage?language=en in the Misc category
You have to thank @SuksAE for the huge amount of work he put into this.

PR #401, opened 2022-06-05
Can storage

SuksAE:
Can storage generator to satisfy issue #330 (https://github.com/florianfesti/boxes/issues/330).

Note: depends on fillHole function, which is a different pull request...


boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+        if self.top_edge != "e":
+            self.corner(90, self.thickness)
+            self.edge(0.5 * self.canDiameter)            
+            self.corner(-90, 0.25 * self.canDiameter)
+        else:
+            self.moveTo(-self.burn, self.canDiameter + self.thickness, -90)
+            self.corner(90, 0.25 * self.canDiameter)
+            self.edge(self.thickness)
+
+        self.edge(0.5 * self.canDiameter - self.thickness)
```
florianfesti, line 34: May be have a look at the `.polyline()` method.

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+            type=ArgparseEdgeType("eEš"), choices=list("eEš"),
+            default="š", help="edge type for bottom edge")
+
+        # Add non default cli params if needed (see argparse std lib)
+        self.argparser.add_argument(
+            "--canDiameter",  action="store", type=float, default=67,
+            help="outer diameter of the cans to be stored (in mm)")
+        self.argparser.add_argument(
+            "--canHight",  action="store", type=float, default=115,
+            help="hight of the cans to be stored (in mm)")
```
florianfesti, line 77: I wonder if there should be a parameter for the number of rows. But having just two as a default is fine for now.

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+            # fill with holes
+            border = [
+                (0, 0), 
+                (self.top_chute_depth, 0), 
+                (self.top_chute_depth, self.width), 
+                (0, self.width),                
+                ]
+            if self.mydebug:
+                self.showBorderPoly(border)
+            if self.fillHoles_fill_pattern != "no fill":
```
florianfesti, line 93: This is a sign of a bad API. `.fillHoles` should just handle the "no fill" option by doing nothing.

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+                    border=border, 
+                    max_radius=self.fillHoles_hole_max_radius, 
+                    hspace=self.fillHoles_space_between_holes, 
+                    bspace=self.fillHoles_space_to_border, 
+                    min_radius=self.fillHoles_hole_min_radius, 
+                    style=self.fillHoles_hole_style,
+                    maxcnt=2000)            
+        
+    def render(self):
+        self.mydebug = False # set to True to enable debugging
```
florianfesti, line 190: Please just use the regular debug parameter. It was added for exactly this purpose.

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+    def cb_top_chute(self, nr):
+        if nr == 0:
+            # fill with holes
+            border = [
+                (0, 0), 
+                (self.top_chute_depth, 0), 
+                (self.top_chute_depth, self.width), 
+                (0, self.width),                
+                ]
+            if self.mydebug:
```
florianfesti, line 91: Why isn't fillHoles just doing the debug output?

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+                self.showBorderPoly(border)
+            if self.fillHoles_fill_pattern != "no fill":
+                self.fillHoles(
+                    pattern="hbar",
```
florianfesti, line 95: The long slots make those chutes very fragile especially of the cans don't rest on the edges of the chute that are connected to the side walls. Beverages are heavy and you may run into all kind of trouble with the chute bending or even breaking.

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+                    hspace=2, 
+                    bspace=2, 
+                    )            
+            
+    def cb_sides(self, nr):
+        if nr == 0:
+            # for debugging only
+            if self.mydebug:
+                # draw orientation points
+                self.hole(0, 0, 1, color=Color.ANNOTATIONS)
```
florianfesti, line 127: This looks like it should be a loop

boxes/generators/can_storage.py
```diff
@@ -0,0 +1,232 @@
...
+            # bottom chute
+            with self.saved_context():
+                self.moveTo(0, 0.5 * self.thickness, self.angle)
+                self.fingerHolesAt(0, 0, self.depth / math.cos(math.radians(self.angle)), 0)
+            # top chute
+            with self.saved_context():
+                self.moveTo(0, self.thickness + self.canDiameter + self.bottom_chute_height + self.top_chute_height + 0.5 * self.thickness, -self.angle)
+                self.fingerHolesAt(0, 0, self.top_chute_depth, 0)
+            # front barrier
+            with self.saved_context():
```
florianfesti, line 160: The front barrier needs to be at least one thickness in to provide the necessary strength. Having an f edge connection here is a bad idea. All the cans are going to roll and hit this with momentum.

florianfesti (commented):
This is a really nice generator and one can clearly see how much work went into this.

But before this can be merged one needs to actually cut it and try with real cans. The cans going around the corner is a tricky thing and small mistakes can easily get them stuck. It is quite possible that more angle surfaces are needed to make that work. Or it is just fine. How knows?

boxes/generators/can_storage.py
SuksAE: ---> fixed in PR #400

boxes/generators/can_storage.py
SuksAE: --> done

boxes/generators/can_storage.py
SuksAE: changed in PR #400

boxes/generators/can_storage.py
SuksAE: --> done

florianfesti:
The stackable edges kinda suggest the dispensers can be stacked. But there is nothing that keeps them in place left to right. Adding stackable edges to the back plate will help with that. On the front one will need small pieces to glue on the inside of the feet to prevent them from slipping.

HaSHsss:
image
This part really upsets me.
It makes it difficult to take cans easily.
You have to pretend to be a crab with claws to hook cans.

SuksAE:
Maybe you want to give it a try?
I hope this doesn`t weaken the design too much...

florianfesti:
May be testing it in a smaller size is quicker and cheaper. May be in a similar size to this:  https://www.thingiverse.com/thing:898062

SuksAE:
Spoiler alert:

CanStorage_AA_cut


CanStorage_AA

Designing a battery holder with CanStorage is not so easy - AA size is a lot harder to manage than 400g cans. But I kind of like the result...

SuksAE:
from my point of view this is ready to be merged into master...

florianfesti:
Sorry for the slow response. I was on vacation, then sick, then sick on vacation. I hope we can wrap this up soon.

Minor thing: The last patch misses changes to `static/samples/samples.sha256`. See https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#provide-photos-for-generators for details. If you don't have the right tooling at hand I can also generate the thumbnails for you and amend the patch.

I am wondering whether we should squash the patches or whether it is worth keeping the different development steps in separate patches. Any opinion on that?

SuksAE:
OK, I missed that with the pictures - if you could handle this, it would be fine.

Regarding the patches - I don't see any benefit in keeping the separate steps. I think the end result is most important outcome, and as there was nothing like a release to the public in between, the intermediate steps can be omitted...

florianfesti:
Merged as 43bd7d12d318e353cc1178f714701a23ef6ed989

Thanks for all the work and patience!

SuksAE:
Final question: when will this be available on www.festi.info/boxes.py?
Is there an update schedule for the web server or how do you handle the updates?

florianfesti:
I am updating the web instance by hand. I am always a bit hesitant to do that and will typically wait until I do have enough time at hand to deal with any issues that may come up. Unfortunately the docker service on this rather old OS isn't too reliable and sometimes needs a love bump.

Anyway, updated.

AshaiRey:
Too ALL worked on this can generator.
Thank you for your hard work to make this possible.
I really appreciate it very much.

PR #402, opened 2022-06-05
Lbrn2 output

SuksAE:
Updated lbrn2 color handling to get a better match with Lightburn layers


boxes/drawing.py
```diff
@@ -733,6 +734,17 @@ class LBRN2Surface(Surface):
         'monospaced' : 'Courier New'
     }
 
+    lbrn2_colors=[
+        0,  # Colors.OUTER_CUT    (BLACK)   --> Lightburn C00 (black)
```
florianfesti, line 738: I wonder if inner and outer cut should be swapped color wise or if they should be re-ordered in the new CutSettings

SuksAE: You are right - the table does the color matching and then there is a separate section where the cutting sequence is defined...

PR #407, opened 2022-06-16
Handle edge

SuksAE:
Added a new edge type (again).
Inspiration came from https://www.thingiverse.com/thing:327393

There is one issue I would need some help:

I need to call the function argparseSections() which is located in __init__.py in edges.py (line 2507). As it is not part of a class, I see no other way as to copy the function to edges.py...

Do you have a better idea? - I guess my Python knowlege is too limited here.

### DIFF
--- a/boxes/edges.py
+++ b/boxes/edges.py
@@ -23,6 +23,39 @@ import abc
 
 from boxes import gears
 
+def argparseSections(s):
+    """
+    Parse sections parameter
+
+    :param s: string to parse
+
+    """
+
+    result = []
+
+    s = re.split(r"\s|:", s)
+
+    try:
+        for part in s:
+            m = re.match(r"^(\d+(\.\d+)?)/(\d+)$", part)
+            if m:
+                n = int(m.group(3))
+                result.extend([float(m.group(1)) / n] * n)
+                continue
+            m = re.match(r"^(\d+(\.\d+)?)\*(\d+)$", part)
+            if m:
+                n = int(m.group(3))
+                result.extend([float(m.group(1))] * n)
+                continue
+            result.append(float(part))
+    except ValueError:
+        raise argparse.ArgumentTypeError("Don't understand sections string")
+
+    if not result:
+        result.append(0.0)
+
+    return result
+
 def getDescriptions():
     d = {edge.char: edge.description for edge in globals().values()
          if inspect.isclass(edge) and issubclass(edge, BaseEdge)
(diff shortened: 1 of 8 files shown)


boxes/edges.py
```diff
@@ -2554,6 +2594,18 @@ def startwidth(self):
         return self.settings.thickness
 
     def __call__(self,  length, **kw):
-        self.fingerHolesAt(0, 0.5*self.settings.thickness, length, 0)
+        self.FingerHolesAt(0, 0.5*self.settings.thickness, length, 0)
```
florianfesti, line 2597: This fails. The f in fingerHolesAt must be lower case.

florianfesti:
There are no inverse finger holes in Boxes.py. I see why there are tempting here but things are not that easy. Have a look at a TypeTray with handle and and "h" or "s" bottom. Unless you print these pieces in place you won't get them assembled. Not even having a inverse finger joint edge makes issues like that impossible. An yes, this means that some boxes can't have a handle this way - at least in most (and most important) configurations.

florianfesti:
Merged the unrelated changes into master already.

SuksAE:
> There are no inverse finger holes in Boxes.py. I see why there are tempting here but things are not that easy. Have a look at a TypeTray with handle and and "h" or "s" bottom. Unless you print these pieces in place you won't get them assembled. Not even having a inverse finger joint edge makes issues like that impossible. An yes, this means that some boxes can't have a handle this way - at least in most (and most important) configurations.

I see your concerns here, but in the parameter description of TypeTray is stated that checking handle makes only sense if F is selcted for bottom edge.

As a mitigation I see these possibilites: 
- force the bottom edge on the front wall to be F in case handle is checked
- force the bottom edge of all walls to be F in case handle is checked
- raise an assertion if handle is checked and bottom edge != F

florianfesti:
What about the opposite way: Switch to a Y edge and make the front bottom edge f if handle is selected.

Generators like the UniversalBox that use `ignore_widths` need a bit of special care. But then you can combine the handle with all possible bottom edges.

florianfesti:
For stackable edges one could even create a piece to glue onto the bottom side...

florianfesti:
There is one more issue with inverted finger holes I missed until now. They may cut off the outer part if the edge is small enough. Normal finger holes are guaranteed to keep the two sides connected (*). This is the final nail in the coffin for the inverted finger holes.

(*) Except you set `surroundingspaces` to zero and the edge is exactly one finger long.

florianfesti:
OK, I fixed a few other issues, deleted the inverse finger joint patch and moved Divider and Type Tray over to a combination of Y and f edges.

Do you want me to force push this here or should I open another PR?

SuksAE:
no, let's keep it simple - use this one...

florianfesti:
From my POV this is now ready for inclusion code wise. Someone(tm) should probably test cut a few of the boxes with the new handle. That won't be me for the next few weeks as I am traveling.

I am not too thrilled about the copy of `argparseSections` but I can clean that up afterwards. Probably by moving all the param handling stuff into it's own module.

SuksAE:
yes, I can cut someting - do you have any special settings in mind?

florianfesti:
May be the new outset variant.

SuksAE:
I am afraid you won't like it:

**It is not enough to have a separate edge for the front edge.** 
Having different edge types won't assemble...

Here a picture with "h" as bottom edge:
TypeTray_HandleFront_nOK1

Also mixing "F" and "f" don't work out:
TypeTray_HandleFront_nOK2


So I had to partly reverted the code **to use inverse finger holes to make it work:**
TypeTray_HandleFront_OK

I know you don't like the inver se finger holes, so I did not include this edge type in the list of regular to be selected edges, but from my point of view there is no way around except trashing the idea of the bottom handle...

SuksAE:
Furthermore, here are some picture of the HandleEdge on top edge:
TypeTray_HandleTop1
TypeTray_HandleTop2
This works out fine...

florianfesti:
> Generators like the UniversalBox that use `ignore_widths` need a bit of special care.

Well, should have taken that...

That's the issue why this doesn't work.  `ignore_widths` here extends the  vertical edges to the bottom. This breaks when not all bottom edges are the same width. Setting all `ignore_widths` to `[]` should fix that. Setting it to `[1]` on one side and [6] on the other and `[]` at the front should be even better.

Let me see if I can fix that...

florianfesti:
Yes, this works just fine. Can't really push this on your current branch though.

florianfesti:
What other generators should get a `handle` parameter?

SuksAE:
> Yes, this works just fine. Can't really push this on your current branch though.

How can I help you here?

SuksAE:
> What other generators should get a `handle` parameter?

I don't know. I guess the boxes for usage in shelfes would benefit.
But I am open for suggestions....

florianfesti:
Well, I'd just rebase the this all into one coherent branch and merge that. I'd remove the [HandleInvHoleEdge](https://github.com/florianfesti/boxes/pull/407/commits/0b341de4feae98658a1d99531e8b179e86ddd663) patch again and merge my fixes.
Unless you have other suggestions...

We can add the handle to more generators later on. I'd rather get this finally in that waiting on more features.

florianfesti:
OK, I hope I didn't mess things up during rebase...

PR #411, opened 2022-07-06
Laptop-stand: create generator

daniel-windsor:
Adding a new generator for a simple laptop stand.  Two triangular pieces fit together to form an X shape on which a laptop can sit.

I haven't yet actually cut a prototype, when I do I'll include a photo in this PR.

In the meantime, any feedback is appreciated :+1:

### DIFF
--- a/boxes/generators/laptopstand.py
+++ b/boxes/generators/laptopstand.py
@@ -65,8 +65,7 @@ class LaptopStand(Boxes):  # Change class name!
     def render(self):
         calcs = self.perform_calculations()
 
-        self.draw_triangle(calcs, False)
-        self.draw_triangle(calcs, True)
+        self.laptopstand_triangles(calcs, move="up")
 
     def perform_calculations(self):
         # a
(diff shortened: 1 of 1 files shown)


florianfesti:
Oh, and after this is finished someone(tm) needs to make a variant that is a guitar stand.

daniel-windsor:
Thanks a lot for your comments!

I've refactored the code to have the same function draw both pieces.  I've left the calcs function in place as I feel it's cleaner to run the calcs once up front instead of running them twice each time the draw triangle function is called.

I've also added feet and rounded the corners on the nub.  Initially I chose not to round them as providing a radius affects the length of that line.  It was tricky to account for the extra length in already complicated calculations.  I've included it in the nub as it uses simple angles, but haven't managed to include it for other corners at this stage.

I also really like the idea of a non-slip notch.  I'll investigate that for a future commit

florianfesti:
OK, I moved the parts a round a bit to make them fit together nicely. From my POV this PR can just be merged. Or should I wait for new features?
I would also squash your two commits into one, if that's fine with you. Any progress on the "actually laser cutting this" front?

daniel-windsor:
Hi @florianfesti , sorry for the delay.  I've checked out your changes and it looks great!

I'd like to hold off on merging until I've actually cut it, just to make sure it works.  I've sent the plan off to be cut so I should be able to confirm in a few days.  I'll add some photos as well if it all works out.  I'll let you know how it goes.

daniel-windsor:
Ok @florianfesti , I've had it printed and it works great.  I haven't managed to take a good photo of it yet, but when I do that I'll attach it to the image thread.  I reckon we're good to merge now :+1:

florianfesti:
Looks pretty nice!

PR #418, opened 2022-07-31
Reworked text appearance in BurnTest generator

SuksAE:
- changed font color to ETCHING, 
- changed font size to depend on tile size, enabling smaller tiles with readable text


florianfesti:
Thanks a lot. There are still many places like this one where things can be polished up a bit. It's great to see someone caring about them.

PR #419, opened 2022-08-10
Dicebox

SniderThanYou:
This PR adds a new generator for a dice box. It is based on the `IntegratedHingeBox`, but requires some extra finger joints for the dice separator panels. I added the finger joint support to three of the `ChestHingeXYZ` edges. If there's a better way to do it, please let me know.

I also updated `regularPolygonHole()` to support rounded corners, because it just looks a little nicer than the sharp corners. I used the `FillTest` generator to verify that the existing functionality is preserved. `FillTest` uses `fillHoles()`, which uses `regularPolygonHole()`

PXL_20220810_021014702

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1231,7 +1231,7 @@ class Boxes:
 
     @restore
     @holeCol
-    def regularPolygonHole(self, x, y, r=0.0, d=0.0, n=6, a=0.0, tabs=0):
+    def regularPolygonHole(self, x, y, r=0.0, d=0.0, n=6, a=0.0, tabs=0, corner_radius=0.0):
         """
         Draw a hole in shape of an n-edged regular polygon
 
@@ -1249,15 +1249,30 @@ class Boxes:
         if n == 0:
             self.hole(x, y, r=r, tabs=tabs)
             return
+ 
         if r < self.burn:
             r = self.burn + 1E-9
         r_ = r - self.burn
+        
+        if corner_radius < self.burn:
+            corner_radius = self.burn
+        cr_ = corner_radius - self.burn
+
+        side_length = 2 * r_ * math.sin(math.pi / n)
+        apothem = r_ * math.cos(math.pi / n)
+        # the corner chord:
+        s = math.sqrt(2 * math.pow(cr_, 2) * (1 - math.cos(2 * math.pi / n)))
+        # the missing portion of the rounded corner:
+        b = math.sin(math.pi / n) / math.sin(2 * math.pi / n) * s
+        # the flat portion of the side:
+        flat_side_length = side_length - 2 * b
+        
         self.moveTo(x, y, a)
         self.moveTo(r_, 0, 90+180/n)
-        l = 2 * r_ * math.sin(math.pi / n)
-        for i in range(n):
-            self.edge(l)
-            self.corner(360/n)
+        self.moveTo(b, 0, 0)
+        for _ in range(n):
+            self.edge(flat_side_length)
+            self.corner(360/n, cr_)
 
     @restore
     @holeCol
--- a/boxes/edges.py
+++ b/boxes/edges.py
@@ -1461,13 +1461,23 @@ Values:
 
  * pin_height : 2.0 : radius of the disc rotating in the hinge (multiples of thickness)
  * hinge_strength : 1.0 : thickness of the arc holding the pin in place (multiples of thickness)
+
+* absolute
+
+ * finger_joints_on_box : False : whether to include finger joints on the edge with the box
+ * finger_joints_on_lid : False : whether to include finger joints on the edge with the lid
 """
 
     relative_params = {
         "pin_height" : 2.0,
         "hinge_strength" : 1.0,
         "play" : 0.1,
-        }
+    }
+    
+    absolute_params = {
+        "finger_joints_on_box" : False,
+        "finger_joints_on_lid" : False,
+    }
 
     def checkValues(self):
         if self.pin_height / self.thickness < 1.2:
@@ -1512,10 +1522,21 @@ class ChestHinge(BaseEdge):
             self.hole(-t, -s-p, p, tabs=4)
             self.rectangularHole(-0.5*t, -s-p-0.5*pinh, t, pinh)
 
-        poly = (0, -180, t, (270, p+s), 0, -90, l+t-p-s)
+        if self.settings.finger_joints_on_box:
+            final_segment = t-s
+            draw_rest_of_edge = lambda : self.edges["F"](l-p)
+        else:
+            final_segment = l+t-p-s
+            draw_rest_of_edge = lambda : None
+
+        poly = (0, -180, t, (270, p+s), 0, -90, final_segment)
+
         if self.reversed:
-            poly = reversed(poly)
-        self.polyline(*poly)
+            draw_rest_of_edge()
+            self.polyline(*reversed(poly))
+        else:
+            self.polyline(*poly)
+            draw_rest_of_edge()
 
     def margin(self):
         if self.reversed:
@@ -1551,10 +1572,22 @@ class ChestHingeTop(ChestHinge):
         p = self.settings.pin_height
         s = self.settings.hinge_strength
         play = self.settings.play
-        poly = (0, -180, t, -180, 0, (-90, p+s+play), 0, 90, l+t-p-s-play)
+
+        if self.settings.finger_joints_on_lid:
+            final_segment = t-s-play
+            draw_rest_of_edge = lambda : self.edges["F"](l-p)
+        else:
+            final_segment = l+t-p-s-play
+            draw_rest_of_edge = lambda : None
+
+        poly = (0, -180, t, -180, 0, (-90, p+s+play), 0, 90, final_segment)
+
         if self.reversed:
-            poly = reversed(poly)
-        self.polyline(*poly)
+            draw_rest_of_edge()
+            self.polyline(*reversed(poly))
+        else:
+            self.polyline(*poly)
+            draw_rest_of_edge()
 
     def startwidth(self):
         if self.reversed:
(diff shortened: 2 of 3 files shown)


florianfesti:
This looks very nice.

But I am currently on vacation. So reviews can take longer or shorter than normal.

SniderThanYou:
@florianfesti take your time :) There is no need to rush.

boxes/edges.py
```diff
@@ -1461,12 +1461,15 @@ class ChestHingeSettings(Settings):
 
  * pin_height : 2.0 : radius of the disc rotating in the hinge (multiples of thickness)
  * hinge_strength : 1.0 : thickness of the arc holding the pin in place (multiples of thickness)
+ * finger_joints_on_{box,lid} : False : whether to include finger joints on the edge with the box or lid
 """
 
     relative_params = {
         "pin_height" : 2.0,
         "hinge_strength" : 1.0,
         "play" : 0.1,
```
florianfesti, line 1470: These need to be absolute_params. There is really no need to scale the bool with thickness

boxes/edges.py
```diff
@@ -1461,12 +1461,15 @@ class ChestHingeSettings(Settings):
 
  * pin_height : 2.0 : radius of the disc rotating in the hinge (multiples of thickness)
  * hinge_strength : 1.0 : thickness of the arc holding the pin in place (multiples of thickness)
+ * finger_joints_on_{box,lid} : False : whether to include finger joints on the edge with the box or lid
```
florianfesti, line 1464: Please make this two separate lines as they are used to generate the descriptions in the UI.

For now there is no way to have hidden parameters that ar enot shown in the UI but are set only by the generators. But this is not the first place where such thing is useful and I might add that feature at some point in the future. Probably somehthing like parameters starting with `_` being hidden.

As they for now can be deselected by the user you can either not render the wall with the hex holes for the side that got unchecked or may be force the setting to `True` in the generator code.

boxes/edges.py
```diff
@@ -1512,10 +1515,21 @@ def __call__(self, l, **kw):
             self.hole(-t, -s-p, p, tabs=4)
             self.rectangularHole(-0.5*t, -s-p-0.5*pinh, t, pinh)
 
-        poly = (0, -180, t, (270, p+s), 0, -90, l+t-p-s)
+        if self.settings.finger_joints_on_box:
+            final_segment = t-s
+            draw_rest_of_edge = lambda : self.edges["F"](l-p)
```
florianfesti, line 1520: I wonder if it is easier to read of you just use a length for the finger joint edge and set that to `0.0` for the normal case. But that's more a question of style.

boxes/generators/dicebox.py
```diff
@@ -0,0 +1,112 @@
...
+        e2 = edges.CompoundEdge(self, "Fe", (h-hy+t, hy-t))
+        e_back = ("F", e1, "F", e2)
+
+        p = self.edges["o"].settings.pin_height
+        e_inner_1 = edges.CompoundEdge(self, "fe", (y-p, p))
+        e_inner_2 = edges.CompoundEdge(self, "ef", (p, y-p))
+        e_inner_topbot = (e_inner_1, "f", e_inner_2, "f")
+
+        self.ctx.save()
+
```
florianfesti, line 96: Mixing `y` and `x` wide parts in the first column is not ideal. This also breaks if y is smaller than x. Either rotate the parts or render them in the right `x` wide column.

boxes/generators/dicebox.py
```diff
@@ -0,0 +1,112 @@
...
+    def diceCB(self):
+        t = self.thickness
+        xc = self.y / 2 - t
+        yc = self.x / 2 - t
+        cr = self.hex_hole_corner_radius
+
+        r = (min(self.x, self.y) - 2 * t) / 2 / math.pi
+        apothem = r * math.sqrt(3) / 2
+
+        # dice
```
florianfesti, line 58: I wonder if this can be done more easier and more generically by using either the new [fillHoles](https://github.com/florianfesti/boxes/blob/master/boxes/__init__.py#L1558) function or one of the old [hex functions](https://github.com/florianfesti/boxes/blob/master/boxes/__init__.py#L1896) probably [hexHolesHex](https://github.com/florianfesti/boxes/blob/master/boxes/__init__.py#L1987).

boxes/__init__.py
```diff
@@ -1249,15 +1249,30 @@ def regularPolygonHole(self, x, y, r=0.0, d=0.0, n=6, a=0.0, tabs=0):
         if n == 0:
             self.hole(x, y, r=r, tabs=tabs)
             return
+ 
         if r < self.burn:
             r = self.burn + 1E-9
         r_ = r - self.burn
+
```
florianfesti, line 1256: I am not 100% decide what I think of that. From an API POV the default radius should be a true `0.0` including the over cuts to make things fit. Otoh these holes are mostly not used for constructive purposes. May we can make the default radius `None` and translate that to `burn` but allow to pass `0.0` for getting the over cuts. This still doesn't play well wit making the radius an UI parameter, tough. May be I am overthinking this...

florianfesti (changes requested):
This looks pretty good already code vise. The few thing that really need changing should be very easy to fix.

SniderThanYou:
@florianfesti Thanks for the review, Florian! I'm looking through your comments now.

boxes/__init__.py
SniderThanYou: @florianfesti It looks like the original author of `regularPolygonalHole` intended this:

`if the radius is less than the burn, then make sure we at least cut _something_, instead of leaving the surface solid`.

However, this seems like a VERY rare edge case, especially since `regularPolygonalHole` is only used by `fillHoles`, and now the dice box. That said, I don't really mind either way. For the purposes of the dice box, no one will ever set the radius to zero; I'm happy to set the radius however you wish. But perhaps that adjustment could be a separate commit, since related to the original implementation, instead of the addition of rounded corners.

The corner radius, on the other hand, could absolutely be zero, with no issue. At the very least, the `+ 1E-9` should be removed from the corner radius burn adjustment. I'll make this change now.

boxes/edges.py
SniderThanYou: Thanks, fixed

boxes/edges.py
SniderThanYou: Thanks, fixed

boxes/generators/dicebox.py
SniderThanYou: Good catch! The defaults create a square box, so I didn't notice it when testing. Fixed.

boxes/edges.py
SniderThanYou: @florianfesti I'll leave this one up to you; this is your code base, and it's important for you to be happy with the style. So if that's how you would prefer, then I'd be happy to make the change.

boxes/generators/dicebox.py
SniderThanYou: @florianfesti It looks like `hexHolesHex` creates circular holes, in a hexagon pattern, in the shape of a hexagon (rather than filling an entire rectangle with holes). But `fillHoles` could do the job more generically. I don't think that it would be easier, though. I need to get to bed, but I can take a look again tomorrow. Thanks again for all the great suggestions!

SniderThanYou: @florianfesti I experimented with `fillHoles()`, and encountered the following problems:

- the distance between the hexagons horizontally is not the same as the distance between them vertically, so it's not as visually appealing
- there's no way to force the radius of the hexagons to be exactly what we want; there's too much internal manipulation of the values
- it really wants to start by putting a hexagon in the top left corner of the area, thus creating a `3-2-3` arrangement, instead of the `2-3-2` arrangement common to dice boxes
- if you add a new parameter to pass through `corner_radius`, then that parameter doesn't make sense for `round` holes, or for scaled-down holes

For now, it seems better to use `regularPolygonalHole()` directly. This was a good experiment though, because it made me realize that my code also had irregular spacing between the x and y directions. I fixed it by just converting to polar coordinates; it was so much easier, and much clearer what we wanted to accomplish.

SniderThanYou:
@florianfesti Back to you, friend! This is ready for a second review.

boxes/generators/dicebox.py
```diff
@@ -0,0 +1,123 @@
...
+        self.ctx.save()
+
+        self.rectangularWall(x, y, "ffff", move="up")
+        self.rectangularWall(x, y, "ffff", move="up")
+        self.rectangularWall(x, h, "FFFF", ignore_widths=[1,2,5,6], move="up")
+        self.rectangularWall(x, h, e_back, move="up")
+        self.rectangularWall(x, hl, "FFFF", ignore_widths=[1,2,5,6], move="up")
+        self.rectangularWall(x, hl-hy2+t, "FFqF", move="up")
+
+        self.ctx.restore()
```
florianfesti, line 115: This one is still the wrong way round and only moves y to the side when it needs to offset the x column.

SniderThanYou: @florianfesti ohhh I get it; I did not really understand the "only" moves before. Fixed, thanks.

florianfesti: Yeah, it isn't that obvious but very convenient once one understands how to use it...

boxes/generators/dicebox.py
florianfesti: Yeah, I guess `fillHoles` - which is really new - needs a bit of polishing to be usable here. It currently lacks the means to align the pattern symmetrically with borders and to control which horizontal offset to start with the hex pattern. That on my todo list as it is necessary to replace the older `hex*` functions 

In the long run I can imagine it being used to have an arbitrary number of holes depending on the size of the box. Just hard coding it to 7 holes is fine for now, though.

boxes/edges.py
florianfesti: Fair enough. Let's not hold this up on questions of style.

florianfesti:
With your permission I would cut the picture to size and add it to the generator. If you want to do that yourself there are brief instructions at https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#provide-photos-for-generators.

florianfesti:
There's an argument to be made that the generator should calculate the size of the hex holes according to size. But if we plan to fit the proper amount of holes at some point, there is no good reason to drop it now.

florianfesti:
Anyway thanks for your work!

SniderThanYou:
> With your permission I would cut the picture to size and add it to the generator. If you want to do that yourself there are brief instructions at https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#provide-photos-for-generators.

Please be my guest. I release the image to the public domain and waive any rights I hold over it.

> There's an argument to be made that the generator should calculate the size of the hex holes according to size. But if we plan to fit the proper amount of holes at some point, there is no good reason to drop it now.

Doesn't it already?

```
apothem = (min(xi, yi) - 4 * t) / 6
r = apothem * 2 / math.sqrt(3)
```

> Anyway thanks for your work!

My pleasure! Thanks for sharing the tool.

HaSHsss:
it's looks like not OK with this fingerhole

image

SniderThanYou:
> it's looks like not OK with this fingerhole
> 
> image

Those tabs fit into the integrated hinge:

PXL_20220818_184628816

PXL_20220818_184639063

florianfesti:
Well, this generator - and the original IntegratedHingeBox - would benefit from labels for the parts.

PR #421, opened 2022-08-13
Fix cursor on main menu, allow build docker on windows

tinohager:


florianfesti:
I wonder if using a bind mount for the sources would be even easier to use. I don't have any experience with docker on Windows but I think that should work even with windows not using bind mounts itself.
That way you can continue to work on your local checkout with your local tools while still testing the changes running locally in your container.

tinohager:
Yes, that would certainly still work. I wanted to test my changes before I make the commit. Since I am otherwise using other programming languages, docker was of course my first choice here. And of course it doesn't help me much if the docker then gets the files from the GitHub repository.

tinohager:
@florianfesti Do you want to implement the change here with the bind mount?

florianfesti:
May be we can just have both variants in the Dockerfile with one being commented out. default can what ever you find more useful for developing under Windows. I really have no experience with that. My gut instinct tells me the bind mount might be easier to use.

florianfesti:
I grabbed the other fixes and merged them into the master branch after cleaning them up a bit. So we make some progress on the PR at least.

tinohager:
@florianfesti i have add the infos for the bind mount in the `Dockerfile` also fix a second issue with python3 files without extension.

tinohager:
Sometimes the request from the web server hangs in Docker only when reloading the page comes. But I don't think this is a problem that only occurs under Windows

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Simple Fedora based Docker container for running Boxes.py
+
+# Build the docker container
+# docker build -t boxes.py .
+
+# Run the docker container
+# docker run -ti -p 4000:8000 boxes.py
+
+# Run the docker container with a bind mount (Windows Command Line)
```
florianfesti, line 9: This looks weird. Why would you only bind mount the scripts directory? The idea for bind mounting is to use the copy of boxes.py outside the container - not just the script. That way any changes to the work dir are visible without rebuilding the container.

tinohager: The problem of bind the full directory structure is this error

> RuntimeError: dictionary changed size during iteration

```
C:\Users\user1\Documents\GitHub\boxes>docker run -ti -p 4000:8000 -v %cd%:/boxes boxes.py
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python3.10/threading.py", line 1009, in _bootstrap_inner
    self.run()
  File "/boxes/scripts/boxesserver", line 75, in run
    if not self.filesOK():
  File "/boxes/scripts/boxesserver", line 64, in filesOK
    self._addModules()
  File "/boxes/scripts/boxesserver", line 55, in _addModules
    for name, module in sys.modules.items():
RuntimeError: dictionary changed size during iteration
BoxesServer serving on host:port :8000...
```

jondkelley:
I get this when trying to run the container :( ```
Traceback (most recent call last):
  File "/boxes/scripts/boxesserver", line 35, in <module>
    import qrcode
ModuleNotFoundError: No module named 'qrcode'
```

florianfesti:
Yes, python3-qrcode is a new dependency. I think I added that to the dockerfile in the repo but you may need to adjust your own copy.

PR #425, opened 2022-08-28
Keyboard interruption for BoxesServer

Ractor:
Added availability to kill server using KeyboardInterrupt (usually CTRL+C). In the master version, the FileChecker thread doesn't stop, this should help it.

(Tested on Windows 10)


florianfesti:
Thanks for the patch!

PR #426, opened 2022-08-28
Coin Showcase/Holder

Electro707:
This PR adds a coin holder, which is a box with an angled coin stand in the middle to showoff a coin.

This is still a WIP, mainly need to clean up the code, document it somewhat, and of course test it out and post pictures of it when I get the chance to laser cut it.

For now I set it as an "unstable" type. I will change this to the "misc" category once completed.

TODO:
- [x] Rename to "CoinDisplay"
- [x] Add picture 
- [x] Move to misc category


florianfesti:
Well, you didn't ask for comments but I'll still give some. These are all just ideas. This can easily get merges without any of these.


It might be worth adding a `bottom_edge` parameter as can be found in other boxes like `ABox`. Especially the stackable edge may look good on this.

While implementing a new Edge type for the double finger joint is technically impressive I wonder if just gluing on the top layer may look better in practice. May be the top lay could even be made a bit bigger to hide to lower layer. Adding the outline as an engraving cut may make the glue up easier.

Technically you do not really need a bottom and can leave the foot open from below. That will make it a bit less sturdy but I guess you won't have too heavy a gold coin. This will require a different set of `bottom_edge`s though.

I guess you like the way this looks right now. But there a few more designs that can be considered. Basically the whole foot box is not really necessary. This of course depends on the use case. If you want to holder to be a bit robust - may be filling the foot with something heavy - the foot is a necessity. If you want to put it behind glass anyway (e.g. for valuable coins) just the vertical walls may be sufficient and may be more minimalist and elegant.

Electro707:
Well I was finally able to test this out. While you are right that the foot box isn't necessary, this specific coin holder was more designed to be inside of an enclosure, with an example below. While the outside enclosure and the coin holder can be two separate pieces, I think it is better to have the bottom plate integrate into the coin holder and that makes it so that the coin holder is centered in the enclosure (instead of deviating a bit due to human error when placing the enclosure on the holder). I can make it so that it is a checkbox whether to generate the outer box with it or not.



This specific generator should probably be renamed into "Valuable Coin Holder" or something similar due to the single-coin and enclosure setup

florianfesti:
Oh, this is assembled the other way round. I thought the coin would be on top instead of inside.

What about  "CoinDisplay"?

Electro707:
Yeah that name works. I will also move it to the "Misc" category unless you think it better fits elsewhere.

I plan on adding the option to have the coin holder support more than 1 coin of various sizes (for example for a grid of coins for demonstration) and the option to disable the outer enclosure. Should I make that part of this PR, or merge this PR and create another one for the upgrades when I get a chance at it?

Electro707:
Done with the basic changes of moving the generator to the Misc category, renamed the class name and file name to `CoinDisplay`, and added a picture for it. This PR can be merged unless it is decided to add the other features mentioned in my last comment to this PR.

florianfesti:
What about adjusting the description to `"""A show case for a single coin"""` ?

florianfesti:
Thanks for the generator. Merged as a single commit with the sample image shrunk to 1200px width

Issue #427, opened 2022-08-31
Accessibility navigation issues for screen reader users


funkonaut:
There are some issues in the web interface that affect non-visual access to the web page. My pull request attempts to solve them by labeling form fields for screen readers to announce the input elements and making clickable headings keyboard navigable. There is some redundancy with aria-labels as well to improve access by announcing the help text as well as the name when form fields are focused.

Using [this accessibility checker.](https://wave.webaim.org/)

funkonaut:
Issues:
-Table elements that wrap the input forms for each box generator announces extra information such as column and row name according to [W3 WAI reference ](https://www.w3.org/WAI/tutorials/tables/) tables should not be used for layout purposes but fixing this would require a large rewrite of the python/html adding role="presentation" tag provides a work around solution 
-Form fields are not labeled and thusly are not read by screenreaders.
-Clickable headings have device dependent event handlers and no tab index, as they are clickable they should be tab navigable and event handlers should be device independent, so functionality is accessible to both mouse and keyboard users

@HaSHsss do you have any additional issues you would like to add?

florianfesti:
#480 should have solved all of the above. Please re-open or file an new issue if there is anything else we can do to improve accessibility.

Closing.

Issue #140, opened 2019-05-10
Pictures missing


THX2112:
Many of the existing pictures are missing, and a lot of boxes don't have pictures or renderings. This makes it hard to figure out what the parameters do, or what the final box should look like without a lot of trial and error.

florianfesti:
Yeah, making good pictures is difficult and a lot of work. I am especially bad at taking photos. So if you could take some and either open an PR or attach them here that would be a great help.

Blancanegra:
I took some photos, please use it at your own convenience.

ElectronicsBox, and opened and closed:
![20191102115526_IMG_8703](https://user-images.githubusercontent.com/32768345/68070189-6dfaed00-fd6b-11e9-93b7-d438372c6b04.JPG)
![20191102114612_IMG_8696](https://user-images.githubusercontent.com/32768345/68070190-6e938380-fd6b-11e9-9bf0-213d4439ba66.JPG)

PaintStorage first version, stacked and with paint cans:
![20191102115936_IMG_8710](https://user-images.githubusercontent.com/32768345/68070199-a5699980-fd6b-11e9-8460-12bed88cd72f.JPG)

LaserHoldFast in action (ugly set but best than nothing):
![20191102114937_IMG_8699](https://user-images.githubusercontent.com/32768345/68070222-e19cfa00-fd6b-11e9-900a-f167b14a942c.JPG)

NotesHolder (default values):
![20191102115634_IMG_8704](https://user-images.githubusercontent.com/32768345/68070232-0db87b00-fd6c-11e9-84c0-44190c5bd6e3.JPG)

florianfesti:
Thank you very much! Added as 9a8ae19 together with some picture of the WineRack I still had somewhere.

florianfesti:
Ok, this ticket is kinda useless unless we actually track the missing pictures. So here's a list:

- BayonetBox
- Rack*Box
- FlexBox5
- HeartBox
- StorageRack
- Slatwall* - need a photo session in our wood workshop
- BurnTest
- ConcaveKnob
- FlexTest - could be rendered image
- FlexTest2 - could be rendered image
- GearBox
- Gears
- LBeam
- Planetary
- Planetary2
- Pulley
- RectangularWall
- WaivyKnob 
- AngledCutJig
- Arcade
- Display
- DrillBox
- Hook
- LaserClamp
- MagazinFile
- MakitaPowerSupply
- OttoLegs - get from Ottobot project page
[...]

Issue #26, opened 2017-02-13
Long descriptions


florianfesti:
Generator should have a longer description than may include assembling instructions or limitations that is shown on the generator page (in contrast to the "description" that shows up in the generator list).

florianfesti:
Infrastructure is now in place. Someone just need to write all the .description texts (markdown is supported)

florianfesti:
OK, just having the ticket here somehow doesn't magically improve the descriptions. Closing.

Issue #25, opened 2017-02-13
Better descriptions for the generators


florianfesti:
Many generators only have very generic descriptions - only repeating the name as a sentence. Descriptions should stay short but add information to the generator name.

florianfesti:
The ticket on its own doesn't seem to do anything useful. Closing.

PR #428, opened 2022-08-31
Fixes Accessibility issues

funkonaut:
There are some issues in the web interface that affect non-visual access to the web page. My pull request attempts to solve them by labeling form fields for screen readers to announce the input elements and making clickable headings keyboard navigable. There is some redundancy with aria-labels as well to improve access by announcing the help text as well as the name when form fields are focused.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -198,7 +198,7 @@ class ArgparseEdgeType:
             ("""<option value="%s"%s>%s</option>""" %
              (e, ' selected="selected"' if e == default else "",
               translate("%s %s" % (e, self.names.get(e, "")))) for e in self.edges))
-        return """<select name="%s" size="1">\n%s</select>\n""" % (name, options)
+        return """<select name="%s" size="1">\n%s</select>\n""" % (name, options) #here
 
     def inx(self, name, viewname, arg):
         return ('        <param name="%s" type="optiongroup" appearance="combo" gui-text="%s" gui-description=%s>\n' %
@@ -218,8 +218,8 @@ class BoolArg:
         if isinstance(default, (str)):
             default = self(default)
         return """<input name="%s" type="hidden" value="0">
-<input name="%s" type="checkbox" value="1"%s>""" % \
-            (name, name, ' checked="checked"' if default else "")
+<input name="%s" id="%s" aria-labeledby="%s %s" type="checkbox" value="1"%s>""" % \
+            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "") #fix here
 
 boolarg = BoolArg()
 
--- a/scripts/boxesserver
+++ b/scripts/boxesserver
@@ -242,10 +242,10 @@ class BServer:
             if len(group._group_actions) == 1 and isinstance(group._group_actions[0], argparse._HelpAction):
                 continue
             prefix = getattr(group, "prefix", None)
-            result.append('''<h3 id="h-%s" class="open" onclick="showHide(%s)">%s</h3>\n<table id="%s">\n''' % (groupid, groupid, _(group.title), groupid))
+            result.append('''<h3 id="h-%s" class="open" onclick="showHide(%s)">%s</h3>\n<table role="presentation" id="%s">\n''' % (groupid, groupid, _(group.title), groupid))
             for a in group._group_actions:
                 if a.dest in ("input", "output"):
-
+                    
                     continue
                 result.append(self.arg2html(a, prefix, defaults, _))
             result.append("</table>")
--- a/scripts/boxesserver
+++ b/scripts/boxesserver
@@ -156,25 +156,25 @@ class BServer:
             viewname = name[len(prefix)+1:]
 
         default = defaults.get(name, None)
-
-        row = """<tr><td>%s</td><td>%%s</td><td>%s</td></tr>\n""" % \
-              (_(viewname), "" if not a.help else markdown.markdown(_(a.help)))
+#fix here?
+        row = """<tr><td id="%s"><label for="%s">%s</label></td><td>%%s</td><td id="%s">%s</td></tr>\n""" % \
+              (name+"_id", name, _(viewname), name+"_description","" if not a.help else markdown.markdown(_(a.help)))
         if (isinstance(a, argparse._StoreAction) and
             hasattr(a.type, "html")):
             input = a.type.html(name, default or a.default, _)
         elif a.dest == "layout":
             val = (default or a.default).split("\n")
-            input = """<textarea name="%s" cols="%s" rows="%s">%s</textarea>""" % \
-                    (name, max((len(l) for l in val))+10, len(val)+1, default or a.default)
+            input = """<textarea name="%s" id="%s" aria-labeledby="%s %s" cols="%s" rows="%s">%s</textarea>""" % \
+                    (name, name, name+"_id", name+"_description", max((len(l) for l in val))+10, len(val)+1, default or a.default)
         elif a.choices:
             options = "\n".join(
                 ("""<option value="%s"%s>%s</option>""" %
                  (e, ' selected="selected"' if (e == (default or a.default)) or (str(e) == str(default or a.default)) else "",
                   _(e)) for e in a.choices))
-            input = """<select name="%s" size="1">\n%s</select>\n""" % (name, options)
+            input = """<select name="%s" id="%s" aria-labeledby="%s %s" size="1" >\n%s</select>\n""" % (name, name, name+"_id", name+"_description", options)
         else:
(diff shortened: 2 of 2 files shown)


florianfesti:
I am all for improving the accessibility. But my HTML foo is barely enough to make the web UI work. So having more detailed explanation what tag are added and why is really needed in the commit messages. It might also help to compile a list in #427 or here to what needs changing.

HaSHsss:
This is my income. The proposed corrections are made to comply with the requirements of HTML 1.1.
But they are too superficial, and will not bring real results.

You also need to consider micro markup, structuring, and much more.

HaSHsss:
I do not want to devalue the work of a person, this is part of the work worth doing, but not at all like that.

funkonaut:
@HaSHsss my apologies I am trying to get the webpage to a point where my students who use screen readers can navigate it as well as my sighted students. Testing my PR with with NVDA, JAWS, VoiceOver, and Windows Narrator all yielded results that  
 allowed for non visual navigation and use of the web page, I am not sure why you say they would not bring real results? 

I would love to learn more about what specifically I did incorrectly, if you could link any resources or sample code I would be happy to update my PR to provide a more robust solution. 

@florianfesti Adding specifics in #427 and updating PR to contain more detail commit messages. Will report back here once I finish up the edits.

scripts/boxesserver
```diff
@@ -241,8 +241,9 @@ class BServer:
                 continue
             if len(group._group_actions) == 1 and isinstance(group._group_actions[0], argparse._HelpAction):
                 continue
-            prefix = getattr(group, "prefix", None)
-            result.append('''<h3 id="h-%s" class="open" onclick="showHide(%s)">%s</h3>\n<table role="presentation" id="%s">\n''' % (groupid, groupid, _(group.title), groupid))
+            prefix = getattr(group, "prefix", None)
```
funkonaut, line 244: Adds event handler for keyboard enter press and tab index to allow for tab navigation on clickable headings and a more device independent event handler.

scripts/boxesserver
```diff
@@ -341,7 +342,7 @@ class BServer:
 """ ]
         for nr, group in enumerate(self.groups):
             result.append(f'''
-<h3 id="h-{nr}" class="open" onclick="showHide('{nr}')"
+<h3 id="h-{nr}" class="open" tabindex="0" onclick="showHide('{nr}')" onkeypress="if(event.keyCode == 13) showHide('{nr}')"
```
funkonaut, line 345: Adds event handler for keyboard enter press and tab index to allow for tab navigation on clickable headings and a more device independent event handler.

scripts/boxesserver
```diff
@@ -242,10 +242,10 @@ class BServer:
             if len(group._group_actions) == 1 and isinstance(group._group_actions[0], argparse._HelpAction):
                 continue
             prefix = getattr(group, "prefix", None)
-            result.append('''<h3 id="h-%s" class="open" onclick="showHide(%s)">%s</h3>\n<table id="%s">\n''' % (groupid, groupid, _(group.title), groupid))
+            result.append('''<h3 id="h-%s" class="open" onclick="showHide(%s)">%s</h3>\n<table role="presentation" id="%s">\n''' % (groupid, groupid, _(group.title), groupid))
```
funkonaut: Adds presentation view to layout table. Probably not the best fix as according to[ W3 WAI best practice is using CSS for visual styling](https://www.w3.org/WAI/tutorials/tables/), but gets it more screen reader friendly without requiring a large rewrite of the web ui.

scripts/boxesserver
```diff
@@ -242,10 +242,10 @@ class BServer:
...
             for a in group._group_actions:
                 if a.dest in ("input", "output"):
-
+
```
funkonaut, line 249: This newline doesn't need to be integrated my bad...

HaSHsss:
I'm sorry, but I did not take into account such a circumstance with the blind.
The fixes you suggested are recommendations for structuring HTML 1.1. In any case, I still think the fixes you suggested are very superficial and we could do better together.
I thought you recommended these Google SEO fixes.
Could you tell us in more detail what functionality your students use.
How would you like to navigate.
I understand how screen reading works, but I've never used it myself.

HaSHsss:
You simply add `labels`for elements so that they can be voiced by means of the system.
But I suspect that since the structure is not explicitly declared, there will still be problems with quick navigation.

florianfesti:
OK, as there is a whole list of issues lets talk about them one by one.

The patch for headings looks pretty good already and works just fine. What would we need to add to announce that the section is closed or expanded. We can easily add something in the showHide() function or in the self.css sections about `h3`.

florianfesti:
I am open to replace the tables by proper CSS layout. May be someone can give me a hint on what to try first. Going with the `role="presentation"` hack is still a lot quicker and what we should do for now. The larger changes can then wait a bit.

florianfesti:
OK, I pushed the later two patches with a bit expanded commit messages. Which mainly leaves the labeling issue. The thing is that these input fields are "labeled" by the text in front of them. I wonder if there is a way to make that accessible to the screen reader users.

funkonaut:
Yeah the commit that wasn't pushed does that with a labelfor tag and id tags, just realized I forget to add it in one place so updating the PR to include everything. Adding comments too.

scripts/boxesserver
```diff
@@ -45,7 +45,6 @@ class FileChecker(threading.Thread):
         super(FileChecker, self).__init__()
         self.checkmodules = checkmodules
         self.timestamps = {}
-        self._stopped = False
```
funkonaut, line 48: sorry I guess I did not integrate your most recent changes sorry about that :/

boxes/__init__.py
```diff
@@ -198,7 +198,7 @@ def html(self, name, default, translate):
             ("""<option value="%s"%s>%s</option>""" %
              (e, ' selected="selected"' if e == default else "",
               translate("%s %s" % (e, self.names.get(e, "")))) for e in self.edges))
-        return """<select name="%s" size="1">\n%s</select>\n""" % (name, options) #here
+        return """<select name="%s" id="%s" aria-labeledby="%s %s" size="1">\n%s</select>\n""" % (name,  name, name+"_id", name+"_description", options)
```
funkonaut, line 201: this adds and id and aria-labeled by that references the id, name, and description from the boxesserver file this allows for screen readers to announce the label to the left of the input and the description to the right of the input for the combo box

boxes/__init__.py
```diff
@@ -219,7 +219,7 @@ def html(self, name, default, _):
             default = self(default)
         return """<input name="%s" type="hidden" value="0">
 <input name="%s" id="%s" aria-labeledby="%s %s" type="checkbox" value="1"%s>""" % \
-            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "") #fix here
+            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "")
```
funkonaut, line 222: got rid of the comment

boxes/__init__.py
```diff
@@ -218,8 +218,8 @@ def html(self, name, default, _):
         if isinstance(default, (str)):
             default = self(default)
         return """<input name="%s" type="hidden" value="0">
-<input name="%s" type="checkbox" value="1"%s>""" % \
-            (name, name, ' checked="checked"' if default else "")
+<input name="%s" id="%s" aria-labeledby="%s %s" type="checkbox" value="1"%s>""" % \
+            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "") #fix here
```
funkonaut: adds associated of label to the left of the input box and description to the right to the input box so screen readers will announce items it does this with an id tag and aria-labeledby tags that reference labelfor tags and id tags there is some redundancy with the aria-labeledby tag referencing the same tag as the label for tag (ie name_id) this is to provide robustness across different screen readers.

funkonaut:
All that is left is to get the web page announcing when the clickable headings are expanded and collapsed, usually would use a menu or button element and the aria-expanded tag but since they are headings with device handlers we will have to go a different route... Going to experiment with a couple of ideas tomorrow

florianfesti:
Another issue with accessibility is of course the often poor descriptions of the generators (and the parameters). Many generators rely on a sample image to give an idea what they are producing and some are even lacking that.

See #25 and #26
#140 is linked from generators that still lack an image. But those won't help screen reader users a whole lot.

If you or your students want to add descriptions they can be just added as comments in those tickets. PRs are ofc also welcome.

florianfesti:
@funkonaut Can you and your students have a look at the new search functionality on the generator selection page, please. I am pretty sure it will need some tweaking to work well with screen readers.

Not sure if it's worth bothering with the gallery (link in the footer).

PR #432, opened 2022-09-12
Optimize image size

tinohager:
Optimize image size with tinypng. Recalculate the samples.sha256 hashes

44MB -> 7MB


tinohager:
As far as I have seen there is a script that generates the thumbs. If desired, we can of course reduce only the large images.

florianfesti:
At first glance this seems too good to be true. But the quality seems to be more than enough.

Having a closer look shows that there are some pictures that can also be shrunk down to the default size of 1200 width: LaserHoldfast.jpg, ElectronicsBox.jpg, UniversalBox.jpg

Do you want to resize and recalculate them from the originals?

Generating the thumbnails with the script results in even smaller results - not in a way that really matters. But converting them from the new versions will keep things a little bit more consistent.

If you have trouble with the script I am happy to amend the commit with the results.

tinohager:
I will adjust the PR this week

tinohager:
Resize main images to 1200px and compress main images

tinohager:
@florianfesti Should anything else be adjusted?

florianfesti:
Rebuild the thumbnails and merged as 0745ec46959835b78828f08de144b1f1b76db9eb. Thanks for all the work. This will surely save cycles on my server and make the UI more responsive.

PR #437, opened 2022-09-30
HalfBox

SuksAE:
A new generator for creating a bookend, a hanging shelf, an angle clamping jig, and so much more...

...and again I have to ask you to update the sha256 codes for the images.


florianfesti:
Nice one. This will be handy for a lot of use cases.

PR #444, opened 2022-10-11
Fix processing of grip_percentage parameter

myokaski:
grip_percentage is initialized with integer zero value but was not divided by 100 when used.
In this small patch division by 100 is added


florianfesti:
Thanks for the fix.

I was hesitant at first as such a change most of the time needs other places to be adjusted, too. But it looks like this is really all that's necessary.

PR #460, opened 2022-12-03
Add edges to polygonWall

### LINKED ISSUE #459: Advice Requested: multiple edge types on polygonWall
PolygonWall currently supports a single edge type and applies it to all edges, but I'd like to apply different edge types to each segment.  

This could be a change to polygonWall (my favorite), a new method polygonWallWithEdges, or just a local method in my generator (my least favorite).  I'm looking for advice on how to approach this and any gotchas that may come up.  

The context for this is that I want to include finger joints on the highlighted line in the following geometry.  I'm currently creating this shape using polygonWall, but I am open to other suggestions.
image

Ranthalion:
Fixes #459 by allowing multiple edge definitions for polygonWall.  Also added function documentation to polygonWall.

The call on line 2710 to _polygonWallExtend may need additional comments since it does not account for multiple edge types.  I'm not entirely sure how to use those to calculate the extents of the shape.

I did spot check the output of console and heart to confirm that this change did not introduce a regression error, but did not attempt multiple variations.  I didn't see any automatic tests, but let me know if there are any e2e or snapshot tests that could be run to confirm behavior.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -219,7 +219,7 @@ class BoolArg:
             default = self(default)
         return """<input name="%s" type="hidden" value="0">
 <input name="%s" id="%s" aria-labeledby="%s %s" type="checkbox" value="1"%s>""" % \
-            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "") 
+            (name, name, name, name+"_id", name+"_description",' checked="checked"' if default else "")
 
 boolarg = BoolArg()
 
@@ -1182,7 +1182,7 @@ class Boxes:
             self.ctx.restore()
             if self.labels and label:
                 self.text(label, x/2, y/2, align="middle center", color=Color.ANNOTATIONS, fontsize=4)
-            self.ctx.stroke() 
+            self.ctx.stroke()
 
         for term in terms:
             if not term in moves:
@@ -1252,11 +1252,11 @@ class Boxes:
         if n == 0:
             self.hole(x, y, r=r, tabs=tabs)
             return
- 
+
         if r < self.burn:
             r = self.burn + 1E-9
         r_ = r - self.burn
-        
+
         if corner_radius < self.burn:
             corner_radius = self.burn
         cr_ = corner_radius - self.burn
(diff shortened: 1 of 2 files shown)


florianfesti:
This patch doesn't really do what it promises as it does not actually allow to pass multiple edge types to `.polygonWall()`. Adding this feature to `.polygonWall()` only doesn't really make sense as it is supposed to work in combination with `.polygonWalls()` which creates the perpendicular walls. This will require a bit of work as it needs to handle straight edges that should probably generate an opening.

Then there is the whole topic of `.polygonWall()` only supporting edge types with a `.startwidth()` and .endwidth()` of zero. One could argue this is out of scope of this PR but it at least needs to be added to the doc string.

One could argue that adding the docstring and adding it to the documentation should be done for `.polygonWall()` and `.polygonWalls()` at once.

This is not something you need to do all on your own  and I am happy to do parts of that myself. Just listing here so it is written down somewhere.

Ranthalion:
@florianfesti Thank you for the feedback.

I'm not sure I understand the comment that the "patch doesn't really do what it promises", unless it is explicitly the polygonWalls miss.  

I was able to successfully use this new implementation to create the geometry shown in the linked issue.
[DisplayShelf lines 83 - 90](https://github.com/Ranthalion/boxes/blob/1f64ba55c1382f93eb393392ea288c70c66a1abf/boxes/generators/displayshelf.py#L83) in my local branch shows the new functionality to use different edge types for each segment.
image

This code is able to generate the desired shape.


I didn't notice any topic where polygonWall() only supports edge types with a .startwidth() and .endwidth()` of zero.  Is there an issue that describes this?  I'm happy to help resolve these issues once I understand them better.

On the other hand, perhaps there is a different approach that I should be taking to generate the desired geometry.  I'm open to that feedback as well.  This is what I'm making.

florianfesti:
Sorry, I miss-read the code. But it breaks when passing a single Edge instance. I fixed that and put the white space changes into a separate patch.

Please test and then this can go in as is. The other issues are much more work and not in the scope of this change.

Ranthalion:
@florianfesti Thanks for helping with this!  I've pulled the changes and have manually tested as well.  I think this PR is good to be merged.

florianfesti:
Thanks for the patch!

PR #466, opened 2022-12-07
DisplayShelf: Implement back and slope

### LINKED ISSUE #464: DisplayShelf Enhancement: Slope the top and include back panel
I would like DisplayShelf to be enhanced with a few new options.
1. ~~Include holes to run a power cord~~
2. Cut the top of the side walls based on the shelf angle.
3. Create a back wall

Ranthalion:
Implement Sloped Sides and Back Wall for DisplayShelf.

Fixes #464

### DIFF
--- a/boxes/generators/displayshelf.py
+++ b/boxes/generators/displayshelf.py
@@ -18,7 +18,7 @@ from boxes import *
 
 class DisplayShelf(Boxes): # change class name here and below
     """Shelf with slanted floors"""
-    
+
     ui_group = "Shelf"
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


florianfesti:
This looks pretty good already. There are just two minor issues:

When both `include_back` and `slope_top` is selected the sides and the back wall both have `f` type edges which won't fit together.
Also the size calculation for the side walls is missing the width of the finger joints. This is a bit easier to spot if you select the `debug` option.

With these two issues solved this PR can be merged.

florianfesti:
OK, converted the back wall to `F` edges and both side wall variants to `f` edges.

florianfesti:
The size calculation is an issue with `polygonWall` that's (sortof) unrelated to this PR. I'll fix that separately.

Issue #36, opened 2017-02-26
Add settings for Bedbolts


florianfesti:
While the library supports "bed bolts" (slots for nuts and screws holding walls together at joints) the generators don't.

miguelmorin:
Hi! Well done with the project. Your presentation has a "Box2.py with bedbolts". I could not find such a file in the repo. So can I produce files for a box with bedbolts, or do I have to wait for this issue to close?

florianfesti:
Box2 got renamed to UniversalBox - which already has a settings page like a cockpit of a 747. So that's probably not the right place to start with this feature.

If you want to toy around with bed bolts you can remove the following line from boxes/generators/closedbox.py:
`d2 = d3 = None`

The issue with this ticket is that I only use ply wood for my boxes which works just fine without bed bolts. And I really don't know what settings one would want/need to make them useful. If you are able to provide some input adding the code may be not that difficult.

florianfesti:
You should at least see some bed bolt slots. I "fixed" the closed box to have them at all the right places - not just somewhere random.

miguelmorin:
Mmm, I have these lines in `boxes/generators/closedbox.py`:
```
        d2 = edges.Bolts(2)
        d3 = edges.Bolts(5)

        #d2 = d3 = None
```
And I see no bed bolts, as in this image where Inkscape emphasizes the anchor points of the path for one of the sides:
<img width="477" alt="no bed bolts in box" src="https://user-images.githubusercontent.com/32396311/43210413-f332ac0a-9026-11e8-9aa4-80f810f2da98.png">

Do you get bed bolts when you comment out line 45?

florianfesti:
Yes.

miguelmorin:
I think I have the wrong setup. If I rename `closedbox.py` to `closedbox2.py` or add a false assertion in the class:
```
class ClosedBox(Boxes):
    """Fully closed box"""

    assert False, "Does the script hit this line?"
```

then the call

```
scripts/boxes closedbox --thickness=5 --output=test.svg --format=svg --x=254 --y=300 --h=100
```

still produces a file, instead of an error. @florianfesti can you tell me your setup?

florianfesti:
Hmm, looks like it imports the files from somewhere else. May be you installation in the the main Python instance. You can do import boxes.generators.closedbox in the Python shell and then print(boxes.generators.closedbox.__file__)

florianfesti:
No. This is probably a coincident that I have not run into this problem.
[...]

PR #468, opened 2022-12-21
Flag to Enable Bed Bolts

xloem:
I enabled bed bolts. What is the origin of that term?

- made `NutHole.sizes` into global `HexSizes` and added shaft diameter values
- added non-metric values to `HexSizes`
- added flags `--bolts` and `--nut_dist` and parsed passed bolt strings
- enabled bed bolts in classes that use them, if `--bolts` is passed

EDIT:
Things that could be immediately done to improve this:
- note: the bed bolt setup has incomplete design; the old code may not reflect a good plan
- [ ] move `bolts` and `nut_dist` to i think `FingerJointSettings`
  - this may also make some `bedBoltSettings` redundant
- [ ] remove `if self.bolts` from generators. pass the d2 and d3 bolt policies and detect whether or not to use them in the edges.
- [ ] move `HexSizes` to its own module
- [ ] add a comment here so Florian finds the changes


florianfesti:
First of all it is great someone is interested in this feature. I added this very early in the life of Boxes.py but never had any use of it myself. As such is is still very rudimentary and has many possible improvements. Also many details are missing a proper plan. So don't try too hard to understand what the code is trying to do.

As for your changes: Moving the bolt data to a global is a step in the right direction. I wonder if we even should move that into its own module.

I am not too keen on adding more global options. `bolts` and `nut_dist` could probably go into `FingerHoleSettings`.

Generators should not need to do `if self.bolts` but the edges should handle that. So the generator would just supply the number/patterns of bolts but not deal with the details. (The current code is fine to play a round and get a better idea how tings should work).

florianfesti:
First of all it is great someone is interested in this feature. I added this very early in the life of Boxes.py but never had any use of it myself. As such is is still very rudimentary and has many possible improvements. Also many details are missing a proper plan. So don't try too hard to understand what the code is trying to do.

As for your changes: Moving the bolt data to a global is a step in the right direction. I wonder if we even should move that into its own module.

I am not too keen on adding more global options. `bolts` and `nut_dist` could probably go into `FingerHoleSettings`.

Generators should not need to do `if self.bolts` but the edges should handle that. So the generator would just supply the number/patterns of bolts but not deal with the details. (The current code is fine to play a round and get a better idea how tings should work).

florianfesti:
Bed bolts is hardware to assemble beds and other furniture. It's a screw that typically goes through a corner post and is anchored in the side board with a nut embedded after some distance.

While this feature here uses "normal" machine screws they are used in the same way as bed bolts. If a native speaker has suggestions for a better name I am really not that attached.

xloem:
I had heard of them as "T-bolts", although I get the impression that they aren't used much anymore.

florianfesti:
If the parameters move to the FingerJointSettings may be just "bolts" is specific enough.

florianfesti:
OK, added at least the patch with the imperial screw sizes. No point for that to rot forever.

florianfesti:
Closing this PR as it needs updating. We still have #36 open.

PR #487, opened 2023-01-06
Create a gridfinity base and holding tray

ccrome:
Hi there, 
  I created gridfinity base and tray holder tray for putting gridfinity holders in.  

Seems to work well. 

The thing I'm not clear on is why do I need an extra space of t/4 in the code to get things to align properly?

specifically here:
```python
    def create_base_plate(self):
        pitch = self.pitch
        nx, ny = self.x, self.y
        t = self.thickness
        opening = self.opening
        offset = t/4 # I don't know why I need to offset the holes by t/4.
        with self.saved_context():
            self.rectangularWall(nx*pitch, ny*pitch, move="left", ignore_widths=[])
            for col in range(nx):
                for row in range(ny):
                    lx = col*pitch+pitch/2+offset
                    ly = row*pitch+pitch/2+offset
                    self.rectangularHole(lx, ly, opening, opening)
        return
```
Let me know what changes I need to make if you'd like to integrate this.

Thanks,
 -Caleb


florianfesti:
You need to use the callback interface to drawn on parts like the rectangularWall: https://florianfesti.github.io/boxes/html/api_parts.html
Then the offset goes away and your code becomes independent of the edge types used around the rectangularWall.

ccrome:
ah ha!  Got it correct with a callback.  Thanks.  Pushed a new update that fixes that issue.

florianfesti:
This looks pretty good already. Just move from ui_group "Unstable" to "Tray"

ccrome:
Okay, updated to 'Tray' and verified that it shows up in 'Trays and Drawer Inserts'

PR #499, opened 2023-01-09
Added option for dividers in display shelf model

patvdleer:
Added an option to add dividers in the shelves of the display shelf model. I have not yet extensively tested this yet. 

Afterthought:
It would be nice if I could make this from thinner material but the code doesn't seem to (easily) allow multiple thickness (or I'm overlooking something)

Examples, 100x100x100 box, (inside), thickness of 3.62 mm:

image

image

### DIFF
--- a/boxes/generators/displayshelf.py
+++ b/boxes/generators/displayshelf.py
@@ -16,10 +16,27 @@
 
 from boxes import *
 
-class DisplayShelf(Boxes): # change class name here and below
+
+class DisplayShelf(Boxes):
     """Shelf with slanted floors"""
 
     ui_group = "Shelf"
+
+    # arguments/properties
+    num: int
+    x: float
+    y: float
+    h: float
+    angle: float
+    thickness: float
+    radians: float
+    sl: float
+    front_wall_height: float
+    include_back: bool
+    slope_top: bool
+    outside: bool
+    num_of_dividers: int
+    divider_wall_height: float
 
     def __init__(self) -> None:
         Boxes.__init__(self)
@@ -28,13 +45,13 @@ class DisplayShelf(Boxes): # change class name here and below
 
         self.buildArgParser(x=400, y=100, h=300, outside=True)
         self.argparser.add_argument(
-            "--num",  action="store", type=int, default=3,
+            "--num", action="store", type=int, default=3,
             help="number of shelves")
         self.argparser.add_argument(
-            "--front_wall_height",  action="store", type=float, default=20.0,
+            "--front_wall_height", action="store", type=float, default=20.0,
             help="height of front walls")
         self.argparser.add_argument(
-            "--angle",  action="store", type=float, default=30.0,
+            "--angle", action="store", type=float, default=30.0,
             help="angle of floors (negative values for slanting backwards)")
         self.argparser.add_argument(
             "--include_back", action="store", type=boolarg, default=False,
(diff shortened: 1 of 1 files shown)


patvdleer:
Laser is currently running but so far so good!

image

florianfesti:
An bit more general variant could be replacing `x` by `sx` as is used in many trays or turn the `x` parameter into sections. See https://florianfesti.github.io/boxes/html/usermanual.html#section-parameters
That way the sections can have individual widths.

florianfesti:
The current code looks like things won't fit if `front_wall_height` and `divider_wall_height` are not the same value as the finger joint of the dividers and the finger holes in the front won't match.

The case with lower dividers is easy to fix. But the interesting case is dividers that are higher. In this case using a CompoundEdge at the front with the lower part "f" and upper part "E" should do the trick.

Just using `front_wall_height` for the divider heights is an easier way out of this if you don't want to put too much more work into the change.

This is a very nice feature both ways.

patvdleer:
> The current code looks like things won't fit if front_wall_height and divider_wall_height are not the same value as the finger joint of the dividers and the finger holes in the front won't match.

I did not consider that, I'll look into it.

In the mean time; I made the display for the (water based) square paint bottles of Revell. 

image

image

patvdleer:
I removed the `divider_wall_height` option temporarily, I'll do a proper fix after some more testing.

florianfesti:
Do you need any help? This is a very nice feature I'd like to see getting finished up and merged.

patvdleer:
Ill fix it asap, got a little sidetracked.

patvdleer:
I fixed 2 out of 3 but Im having some issues with the CompoundEdge for when the divider wall is higher than the front wall.

I think I need to do the following, is this correct?
1. determine the finger length
2. calculate the amount of fingers
3. create a list of the fingers and dimensions

```python
if self.divider_wall_height > self.front_wall_height:
    length = self.front_wall_height
    fingers = 2
    f_lengths = []
    edges_ = [
        "f",
        edges.CompoundEdge(self, "".join(["f" * fingers]), f_lengths),
        "e",
        "e"
    ]
```

florianfesti:
No, that would be horrible. The Boxes API was written by a lazy person fora lazy person! You only need to do

```
edges_ = [
        "f",
        edges.CompoundEdge(self, "fe", [self.front_wall_height, self.divider_height - self.front_wall_height]),
        "e",
        "e"
    ]
```

It is only important that the `f` edge and the matching `F` edge or `fingerHoles` have the same length (and settings). The edges take care of the details and make sure they fit together.

patvdleer:
image
image
image

florianfesti:
Oh, sorry. This ofc has to be `"fE"` so the wall protrudes to the same level as the front wall.

patvdleer:
So like this? It's fine by me either way, any preference? 

image

florianfesti:
Currently I don't really have a preference. Both are just fine with me.

To me it looks like this is ready to be merged. I will probably change the num_dividers and replace it with x supporting sections (similar to the TypeTray but only in one direction). But I am fine with merging this now unless you want to try that yourself or want to do anything else first.

patvdleer:
No I'm fine with the current state and I'm happy with my new storage rack.

florianfesti:
OK, great. Squashed and merged.

PR #504, opened 2023-01-17
Fix boxesserver

ccrome:
commit 0a8dcf0e37880c056d85523cdea97d8c8b166c6e caused the server to *only* respond to localhost by default.  This causes the Docker to not respond properly.  The old server had the default server name as '', but the new one set it to 'localhost'.

This fixes that problem.

In addition, This commit changes the docker file so that you can specify which git repo and commit to check out from.  This allows trivial testing of other forks/branches/users.  The default branch is master and default git repo is Florian's, but you  can now specify new commits and/or branches to clone with

```docker build --build-arg BUILD_BRANCH=<branch-or-commit> --build-arg BUILD_REPO=<my-repo> ...```

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -17,7 +17,10 @@ FROM fedora:latest
 RUN dnf install -y git-core python3-markdown python3-setuptools python3-affine python3-shapely pstoedit && dnf clean all
 
 # Get Boxes.py sources to /boxes
-RUN git clone --depth 1 -b master https://github.com/florianfesti/boxes.git
+ARG BUILD_BRANCH=master
+ARG BUILD_REPO=https://github.com/florianfesti/boxes.git
+RUN git clone ${BUILD_REPO} && cd boxes && git checkout ${BUILD_BRANCH}
+RUN chmod +x /boxes/scripts/boxesserver
 
 # Internal port used
 EXPOSE 8000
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/Dockerfile [hunk 1]
```diff
@@ -17,7 +17,10 @@ FROM fedora:latest
 RUN dnf install -y git-core python3-markdown python3-setuptools python3-affine python3-shapely pstoedit && dnf clean all
 
 # Get Boxes.py sources to /boxes
-RUN git clone --depth 1 -b master https://github.com/florianfesti/boxes.git
+ARG BUILD_BRANCH=master
+ARG BUILD_REPO=https://github.com/florianfesti/boxes.git
+RUN git clone ${BUILD_REPO} && cd boxes && git checkout ${BUILD_BRANCH}
```
Rotzbua, line 22: Before only a shallow clone was done with the necessary branch. Now the complete repo is cloned. This increases traffic and image size. Is this intended?

ccrome: Sorry, my fault for merging two separate things into one PR.  

This is because `git clone --depth 1` doesn't work with a SHA tag.  I wanted to be able to easily build the docker for literally any commit, and since `--depth 1` doesn't allow for that, I changed it to a `clone` followed by a `checkout`

I'm sure there's a better way to do it, this was just most expedient, and it was late :-)

Is there a better way you know of?

ccrome: BTW, the reason I wanted the docker image to be built with a specific commit, rather than branch/tag was to find exactly which commit was breaking the docker image.  That is why these happen to be lumped together.  I was doing a manual `git bisect`, passing in each SHA tag to the docker build image.

PR #509, opened 2023-01-19
Add files via upload

Tom-Obvious:
Added a simple, but not yet perfekt, german translation


florianfesti:
Thanks for all the work!

PR #511, opened 2023-01-21
Attempt at adding an additional row to paint storage rack

patvdleer:
I added an additional bottom plate with holes for taller stuff like (modelmaking) paint brushes.

Somehow it wasn't as easy as duplicating the line for the above finger holes. Not sure if I'm missing something or there is a bug.

Swapping these lines below will cause the rectangularHole and the fingerHoles to be rotated 90 degrees.
```python
            wall_callbacks = [
                lambda: self.rectangularHole(h / 3, (x / 2.0) - t, hh, hx, r=hr),
                lambda: self.fingerHolesAt(0, self.canheight / 3, x, 0),
            ]
```

### DIFF
--- a/boxes/generators/paintbox.py
+++ b/boxes/generators/paintbox.py
@@ -21,7 +21,15 @@ class PaintStorage(Boxes):
     """Stackable storage for hobby paint or other things"""
 
     webinterface = True
-    ui_group = "Shelf" # see ./__init__.py for names
+    ui_group = "Shelf"  # see ./__init__.py for names
+
+    canheight: int
+    candiameter: int
+    minspace: int
+    additional_bottom: int
+    additional_top: int
+    hexpattern: bool
+    drawer: bool
 
     def __init__(self) -> None:
         Boxes.__init__(self)
@@ -41,6 +49,12 @@ class PaintStorage(Boxes):
             "--minspace", action="store", type=int, default=10,
             help="Minimum space between the paintcans")
         self.argparser.add_argument(
+            "--additional_bottom", action="store", type=boolarg, default=False,
+            help="Additional bottom/floor with holes the paintcans go through")
+        self.argparser.add_argument(
+            "--additional_top", action="store", type=boolarg, default=False,
+            help="Additional top/floor with holes the paintcans go through")
+        self.argparser.add_argument(
             "--hexpattern", action="store", type=boolarg, default=False,
             help="Use hexagonal arrangement for the holes instead of orthogonal")
         self.argparser.add_argument(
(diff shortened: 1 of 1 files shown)


patvdleer:
FF'ed the master

patvdleer:
Please review, I think I'm missing something, the holes keep rotating

```python
            # lambda: self.fingerHolesAt(0, self.canheight / 3, x, 0)
            # ToDo: this is horrible, I have no clue why the above lambda isn't set to 90 degrees
            if self.additional_bottom:
                wall_callbacks.append(lambda: self.fingerHolesAt((self.canheight / 6)-(t/2), -t, x, 90))
            if self.additional_top:
                if self.additional_bottom:
                    wall_callbacks.append(lambda: self.fingerHolesAt(0, (self.canheight / 6)-t, x, 0))
                else:
                    wall_callbacks.append(lambda: self.fingerHolesAt(self.canheight - ((self.canheight / 6) + t), -t, x, 90))
```

florianfesti:
Yeah, that's not how this works. There is a callback for each side. You want to draw all the fingerholes from the first callback. May be the best way is to not use a lambda expression but a function or method. There is a hack to still use lambda expressions and just use a list of calls within. This is kinda ugly but used at some places.

See also https://florianfesti.github.io/boxes/html/api_parts.html#the-callback-parameter

and https://github.com/florianfesti/boxes/blob/master/boxes/generators/rollholder.py#L76 for an example of an callback method.

patvdleer:
@florianfesti much better!

```python
    def sidesCb(self):
        x, y = self.x, self.y
        t = self.thickness

        stack = self.edges['s'].settings
        h = self.canheight - stack.height - stack.holedistance + t
        hx = 1 / 2. * x
        hh = h / 4.
        hr = min(hx, hh) / 2

        if not self.drawer:
            self.rectangularHole(h / 3, (x / 2.0) - t, hh, hx, r=hr)
            self.fingerHolesAt(((self.canheight/3)*2)-t*2, -t, x, 90)

            if self.additional_bottom:
                self.fingerHolesAt((self.canheight / 6) - (t / 2), -t, x, 90)
            if self.additional_top:
                self.fingerHolesAt(self.canheight - ((self.canheight / 6) + t), -t, x, 90)
        else:
            self.rectangularHole(h / 3, (x / 2.0) - t, hh, hx, r=hr)
```

florianfesti:
Just place all the horizontal layers in one column. They have all the same width and create a nice rectangle.

Also rebase and squash your commits. The "minor fix" can stay separately everything else should go into one commit. If you are not that fluent in git I can also do that after the placement got fixed.

patvdleer:
> Also rebase and squash your commits. The "minor fix" can stay separately everything else should go into one commit. If you are not that fluent in git I can also do that after the placement got fixed.

Rebase done via github, I mainly use Gitlab which gives you the option to squash them in the pull/merge request. I'll fix it manually tonight when I get home.

PR #520, opened 2023-01-24
Added GearBox image and thumbnail

ccrome:


florianfesti:
Thanks for the image!

PR #529, opened 2023-01-31
Do traylayout in one step

ccrome:
I have created a gridfinity generator from traylayout.  But in the process I added a bit of javascript so that TrayLayout can be done in a single step.  

Here's an example interaction: https://youtu.be/-FVWVScKaLA

Let me know if you think that makes sense and I can fix it up (for example, the layout i always auto-generated right now, even when it shouldn't be, like with 'save to url'.  It should also prompt the user to regenerate if it's been hand modified).

Also, I deleted the special TrayLayout handling code in boxesserver and other places.

If you don't like it for TrayLayout, that's fine -- though I think i'll keep it for my GridfinityTrayLayout (not quite ready, but works great)

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -285,6 +285,7 @@ class Boxes:
 
     webinterface = True
     ui_group = "Misc"
+    UI = ""
 
     description: str = ""  # Markdown syntax is supported
 
--- a/boxes/generators/gridfinitytraylayout.py
+++ b/boxes/generators/gridfinitytraylayout.py
@@ -1,10 +1,10 @@
 import boxes
 from boxes import Boxes
-from boxes.generators.traylayout import TrayLayout, TrayLayout2
+from boxes.generators.traylayout import TrayLayout
 from boxes.Color import Color
 from boxes import restore
 
-class GridfinityTrayLayout(TrayLayout2):
+class GridfinityTrayLayout(TrayLayout):
     """A Gridfinity Tray Generator based on TrayLayout"""
 
     description = """
--- a/boxes/generators/traylayout.py
+++ b/boxes/generators/traylayout.py
@@ -19,7 +19,7 @@ import boxes
 from boxes import *
 
 
-class TrayLayout(Boxes):
+class TrayLayoutFile(Boxes):
     """Generate a layout file for a typetray."""
     # This class generates the skeleton text file that can then be edited
     # to describe the actual box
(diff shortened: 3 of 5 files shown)


florianfesti:
I wonder whether moving to `sx` and `sy` parameters and getting rid of the measurements in the layout completely is an even better solution.

Getting rid of all the special two stage handling is very tempting no matter what.

scripts/boxes2inkscape
```diff
@@ -140,9 +140,6 @@ class Boxes2INX:
 
     def writeAllINX(self, path):
         for name, box in self.boxes.items():
```
florianfesti, line 142: I doubt we can do this. Well, Inkscape users could type in the whole layout but that's not really practical.

florianfesti:
I guess we can't get rid of the 2 step process for the command line interface - if we want to keep supporting that. Otoh that's what the currently worked on smoke test is using (and also having trouble with the TrayLayout generator).

scripts/boxes2inkscape
ccrome: > I doubt we can do this. Well, Inkscape users could type in the whole layout but that's not really practical.

I've never used the inkscape plugin, so I don't know how it would work.

> I guess we can't get rid of the 2 step process for the command line interface - if we want to keep supporting that. Otoh that's what the currently worked on smoke test is using (and also having trouble with the TrayLayout generator).

But for the command-line, yes it would still be a 2 step process, but I'm not sure it needs to have special code handling.  Could it work like:
```
$ boxes TrayLayoutGenerator --x 50 --y100 --nx 6 --ny 4  -o mylayout.txt
# Edit mylayout.txt
$ boxes TrayLayoutGenerator --input mylayout.txt  -o mybox.svg
```

perhaps something similar for inkscape?  Having specifically named generators in the scripts directory feels wrong to me. :-)

florianfesti:
OK, this is not complete yet and mostly untested and may break the `GridfinityLayout`...

Moved this to the new `sx`, `sy` code. Dropped most of the overly eager move toward only one generator. Still needs  a solution for the no longer supplied `webargs` argument.

Also someone should have a look over my changed of the js code. Not sure if there is better error handling needed or if this just fails gracefully enough already.

florianfesti:
Someone remind me to set the author back to @ccrome before merging...

ccrome:
I'll check it all out, but probably not until the weekend

florianfesti:
OK, squashed into one patch and added proper commit message.

ccrome:
Sorry I've been awol for the last couple weeks.  Thanks for fixing it up!  I'm playing with it now, and it seems to work well.

PR #530, opened 2023-02-03
Add search box

ccrome:
I think this search box works reasonably well, and doesn't look horrid.  Maybe best to put the search bar somewhere else?

Here's a screenshot of how it works.  It auto-filters and displays only the generators that contain the search string. 

Should this be a regex search?  Or just a regular search?  Currently, it's just a regular 'if generator name contains substring' kind of search.

image

As soon as you hit the little 'x' or remove everything from the search box, it reverts back to the original expand/collapse behavior.

### DIFF
--- a/scripts/boxesserver
+++ b/scripts/boxesserver
@@ -303,9 +303,16 @@ class BServer:
 <div style="width: 25%; float: left;">
 <img alt="self-Logo" src="{self.static_url}/boxes-logo.svg" width="250">
 </div>
+
 <div>
+
 <div class="clear"></div>
 <hr>
+
+<div class="search"><form onkeydown="return event.key != 'Enter';" onkeyup=filterSearchItems(); return event.key;">
+\U0001f50d <input autocomplete="off" type="search" onsearch="filterSearchItems();" name="search" id="search"></input>
+</form></div>
+
 <div class="menu" style="width: 100%">
 <img style="width: 200px;" id="sample-preview" src="{self.static_url}/nothing.png" alt="">
 """]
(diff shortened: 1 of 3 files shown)


ccrome:
I guess in an ideal world, it would only expand the categories that actually contain items.  Or maybe all the category titles get hidden, and only the generators themselves are shown?

florianfesti:
There are currently a few issues with this:

First the atrocious look. I am not sure if moving the input to the right solves it this time. I can give that a try but my GUI foo is also pretty limited

What's the license of that search icon?

The search is limited to the (English) generator ID. That will get very weird in other languages. It also would be cool if it matched the short descriptions.


On the plus side I don't think we need any more advanced search patterns. The data we are searching through is very limited and people can just search for one general term and look through the results. This is not the internet that returns  ten million results.

Otherwise this is a very nice feature. Even I catch myself looking for generators from time to time.

florianfesti:
Still not a beauty but a bit more tolerable. Also searching the text of the whole line now. So it work both on the translated name and description but also still on the English class name.

florianfesti:
Replaced the search icon by the search unicode character. And squashed the changes.

florianfesti:
Guess it is good enough to merge now. May be some one with more graphical skills than we two can improve it in the future.

Anyway, very cool features! Thanks for the work!

ccrome:
> Guess it is good enough to merge now. May be some one with more graphical skills than we two can improve it in the future.
> 
> 
> 
> Anyway, very cool features! Thanks for the work!

Awesome.  Thanks for fixing up my lousy work ;-)

I've been swamped with work, so haven't had much time to play.  Sorry that I'm 100% unskilled in design/html/css.

florianfesti:
Well, that's the beauty of open source projects. Not everyone has to do everything.

PR #531, opened 2023-02-03
Add a gallery page to boxesserver

ccrome:
This adds an image gallery to the server so you can visually search through all available generators and just click on the image of the generator you want.

Here's what it looks like:

image

image


ccrome:
latest push includes placholder images
image

florianfesti:
Technically we already have https://florianfesti.github.io/boxes/html/generators.html but this is ofc much more useful.

There are a few things to improve though. Pictures should all fit into 200x200 pixels and need to be scaled if necessary. I played around a bit with CSS attributes but could not quite get it to work.

This "Needs image" thing needs to go. The right thing here is to show the generator name.

I have a patch for showing the generator name and `__doc__` as a mouse over. This breaks for the `GridfinityBase` as it has a link in the __doc__ string but that can easily move to the `.description`

With the two issues above fixed this can go in.

florianfesti:
OK, I think I got it. From my POV this is now ready to be merged after squashing the commits

ccrome:
Ah! so much better.

squashed and pushed.

florianfesti:
Ok, added i18n to the generator names and descriptions and gallery now also uses the cache.

I also added headings for the generator groups to ad a bit of a structure to the page.

PR #533, opened 2023-02-05
Create a GridfinityTrayLayout object.

ccrome:
This is based on the TrayLayout2, but it designed to just give exactly the correct size boxes to fit into a gridfinity system.

It has a modified TrayLayout setup, so it works in one step.  It uses javascript to auto-generate the grid, which means it doesn't currently work for Inkscape or command line.

Any further thoughts on doing TrayLayout in one step?  This is obviously related to  https://github.com/florianfesti/boxes/pull/529.

This GridfinityTrayLayout is useful as-is, which is why I'm submitting it, but I understand it's maybe not really ready for prime time.   

Also, sorry for the lousy photo -- I'll fix it if we think it's close to ready to merge.

You can try it live at https://boxes.crome.org/GridfinityTrayLayout


tiberiuichim:
The layout looks awesome, it would have save me a lot of time if I would have seen this last week.

One nitpick, there's a setting "inner height of inner walls in mm (unless outside selected)(leave to zero for same as outer walls)", but the `outside` checkbox is not available.

ccrome:
> The layout looks awesome, it would have save me a lot of time if I would have seen this last week.
> 
> One nitpick, there's a setting "inner height of inner walls in mm (unless outside selected)(leave to zero for same as outer walls)", but the `outside` checkbox is not available.

Thanks, I haven't gotten back to it for a week or two so I'm glad it works for you.  Yes, it's just a documentation issue, I need to change that message, since it's *always* outside  because the exterior dimensions of the gridfinity tray is what matters.

ccrome:
> > The layout looks awesome, it would have save me a lot of time if I would have seen this last week.
> > One nitpick, there's a setting "inner height of inner walls in mm (unless outside selected)(leave to zero for same as outer walls)", but the `outside` checkbox is not available.

Okay, fixed this help message

boxes/generators/traylayout.py
```diff
@@ -97,15 +97,8 @@ def __init__(self, input=None, webargs=False) -> None:
...
-        if not webargs:
-            self.argparser.add_argument(
-                "--input", action="store", type=argparse.FileType('r'),
-                default='traylayout.txt',
-                help="layout file")
-            self.layout = None
-        else:
-            self.argparser.add_argument(
-                "--layout", action="store", type=str, default="")
+        self.argparser.add_argument(
```
florianfesti, line 100: This breaks the CLI interface for the TrayLayout.

boxes/generators/traylayout.py
```diff
@@ -359,6 +352,7 @@ def render(self):
                             self.edgeAt("e", posx + t + b, posy - t, t, 90)
                 elif x < lx and self.floors[y][x]:
                     # Left edge
+                    print(e, posx, b, posy)
```
florianfesti, line 355: left over debug code

florianfesti:
OK, fixed the above issues and some other minor nitpicks:

removed unnecessary `baseplate_callback` parameter. Removed more print statements and commented out code.

florianfesti:
There is probably a bit more polishing to do wrt CLI and Inkscape. But I'll do that when looking at the TrayLayout PR.

florianfesti:
hmmm, isn't the engraving for the feet on the wrong side of the base plate?

ccrome:
Haha, yes, I carve it upside down.  I didn't know how to do it otherwise.  Is there some way to indicate "bottom"?

ccrome:
Oh!  I haven't made an asymmetric one yet.  The bottom needs to have the option of being mirrored in case it's asymmetric.

florianfesti:
Well, proper parts can just use `move="mirror"`. Unfortunately TrayLayout is a bit archaic. It needs some love for the one-step-rendering anyway. I'll look into that but it may take a few days.

PR #540, opened 2023-02-18
Add photos for console generator (arcade stick)

MINDoSOFT:
Hi,
I've added some photos for the Console box generator.

It is an arcade stick. The buttons, joystick and pcb, were bought from [ultimarc](https://www.ultimarc.com/).

Used an EPILOG FUSION PRO LASER at my[ local makerspace](https://makerspace.cyens.org.cy/equipment/)
to cut it.

Tried to optimise the images as much as possible.

Thanks you very much for sharing your work !


florianfesti:
Did shrink and compact the images and added them to the description of the Console generator so they actually show up somewhere.

PR #543, opened 2023-02-26
New Generator for Deckbox

GagamelML:
HI :)
Nice project, i wrote a new generator for a deckbox (for Trading Card Game Decks, such as Magic the Gathering).
I have never contributed to any open source project, so please feel free to comment and critizise and also tell me if this is how it works. The generator is based on the existing cardbox, that i rewrote to fit other needs. Mostly, the slider now goes to the side rather than the front.
Cheers 
Manuel

### DIFF
--- a/boxes/generators/cardbox.py
+++ b/boxes/generators/cardbox.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # Copyright (C) 2013-2014 Florian Festi
 # Copyright (C) 2018 jens persson <jens@persson.cx>
+# Copyright (C) 2023 Manuel Lohoff
 #
 #   This program is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
@@ -15,7 +16,7 @@
 #   You should have received a copy of the GNU General Public License
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from boxes import edges, Boxes
+from boxes import edges, Boxes, BoolArg
 
 
 class InsetEdgeSettings(edges.Settings):
@@ -42,13 +43,14 @@ class FingerHoleEdgeSettings(edges.Settings):
     """Settings for FingerHoleEdge"""
     absolute_params = {
         "wallheight": 0,
+        "fingerholedepth": 0,
     }
 
 
 class FingerHoleEdge(edges.BaseEdge):
     """An edge with room to get your fingers around cards"""
     def __call__(self, length, **kw):
-        depth = self.settings.wallheight-self.thickness-10
+        depth = self.settings.fingerholedepth-10
         self.edge(length/2-10, tabs=2)
         self.corner(90)
         self.edge(depth, tabs=2)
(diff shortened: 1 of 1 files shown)


florianfesti:
Yes, this is how this works.

Copying another generator is a good way to get started and the result looks pretty nice.
For merging this I am wondering whether we should merge the changes back into `CardBox` and just allow choosing whether the lid slides from the front or from the side. That would reduce the amount of copied code quite a bit.

If you are not comfortable in doing such a chainsaw operation I can do that, also.

Having an example picture for the new generator would also be nice. There is [documentation on how to do that](https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#provide-photos-for-generators) but you can also just attach it here.

GagamelML:
Merging the two sounds like a good idea. I will try to do it myself, but there are some details to get right:
The orientation of the cards changed, with (cardwidth -> stackhight), (cardhight->cardwidth), (h -> cardhight). For the most general usability, the input variables would need to be renamed such that they no longer depend on card orientation. 
Following the convention i see in other generators, i would suggest (h -> compartment_h), (cardwidth -> compartment_x) and (cardhight -> compartment_y). 
Does that sound reasonable or is it too confusing?

photo_2023-03-05_00-37-52
photo_2023-03-05_00-37-55
photo_2023-03-05_00-37-56

jondkelley:
This design is awesome and I hope you get time to get this PR accepted.

florianfesti:
Yeah, having generic names is a good idea. While it is easier for a specific use case to have specialized names many people will use the generators for different things.

There is also the option to move to an `sx` parameter and get rid of the `num` parameter to allow for arbitrary compartment widths. Have a look at the SlantedTray for an similar example

GagamelML:
I tried to implement all the ideas discussed, details are available in the commits descriptions. The new version is more generic and fully covers all options from the old "Cardbox" as well as the "Deckbox".

I did a decent amount of testing, but this is now too large now for me to do thorough testing by myself on all available options by myself. I therefore kept the name "Deckbox", because it would allow for parallel release to the existing "Cardbox" until more people tried it and confirm that it is bug free.

florianfesti:
This looks pretty nice already. Question is now whether to rename this back to CardBox or to keep it as a separate box generator. I am tending to the former but I am open to arguments.

Patches ofc still need to be squashed and rebased onto master.

GagamelML:
Okay, i updated to so that it has no more conflicts. However, i think i messed up the rebasing and squashing. Any tips here?

With the new script, keeping both permanently would be redundand. I suggest we extend the description such that the different intended options are visible (including the original one). Maybe we can even include some default parameters for the different applications.

florianfesti:
Cleaned up the branches and squashed everything into one commit. Added a commit to rename the generator back to CardBox. I'd squash these two, too, before merging and adjust the commit message.

I personally don't use git merge if I can avoid it. Instead I rebase and push with `--force`. This works better if you use a feature branch for everything. As it doesn't mess up your master.

florianfesti:
OK, this has set here for long enough. Thanks for all your work!

PR #545, opened 2023-03-09
qr_code fixes

ccrome:
* use `short_url` on standalone (button) QR code generation
* filter out redundant arguments to further shorten the URL.  Current list to filter are the `format`, and the `qr_code` arguments

I also noticed that when the filter_url` function is called the 'qr_code' argument is in twice, as in, 

```
'qr_code' = 0, `qr_code` = 1
```

not sure why that is, but they both get filtered with this patch.

### DIFF
--- a/scripts/boxesserver
+++ b/scripts/boxesserver
@@ -90,11 +90,17 @@ def filter_url(url, non_default_args):
         return ''
     args = args.split('&')
     new_args = []
+    args_to_ignore = ["qr_code", "format"]
     for arg in args:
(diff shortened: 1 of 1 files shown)


florianfesti:
These double arguments are an artifact of how the checkboxes work in HTML. If not checked they don't send anything. So there is a hidden input that add the `=0` version that is then "over written" by the `=1` if the box is checked. As we use `argparse` for handling those parameters and they may have  default of 1 we need to pass 0 if the checkbox is not ticked.

PR #550, opened 2023-04-08
New genarators

J-Waal:
CompartmentBox
Desksign
Piratechest
Shoe


florianfesti:
Very nice generators. Any chance to get pictures with them (as soon as they are final)? I am currently trying to cut down on the number of generators without example pics. No need to do the full shebang (https://github.com/florianfesti/boxes/blob/master/CONTRIBUTING.rst#provide-photos-for-generators), just attaching the pictures here is good enough.

While not necessary having separate pull requests for each new generators will allow those that are ready to go in first while others are still in the works of getting out of "Unstable".

The Shoe is really neat. I wonder if the tip could also be rounded over,  to get an even more shoe-like look. Both horizontal or vertical bending would probably work and look well.

The PirateChest looks pretty much ready to go. There already is the normal "Chest" lid with flex. But that lacks both the hinge and also is too tall for my taste. I wonder if a mix of both could also look good. So basically like the PirateChest but the first and last slanted walls are connected with flex.  Just some random idea. Not any reason to not just merge the PirateChest as is.

The DeskSign is simple enough to just go in. Feet e.g. with the stackable edge could make it a bit more fancy but I don't know if that is really necessary.

Wrt the CompartmentBox: I am currently reworking the lid situation. The TypeTray just got options to get lids. But there is no support for this sliding lid yet. I kinda agree that this type of lid is very desirable on all kind of trays and boxes. I might come up with a general solution for that. But I don't have a good plan right now.

J-Waal:
I have pictures for Shoe and DeskSign. For PirateChest I can send the image next week because this box is used is stored somewhere else.

For DeskSign I am not sure if stackable edges may Couse problems with the angled piece. By combining both triangular pieces, less material is wasted. This makes it impossible to mirror only one of the triangles, so if on one side to top of the material if facing outside, the top feces inside on the other side.

For Shoe, I was conservative with the bend radius. Adding more bends at the front would require more space. This was based on a paper craft template, with small modifications for laser cutting.

I wanted to do PirateChest with flex to create the rounded lid, but had problems figuring out how parts had to be attached. Some extra strength comes from gluing the all parts of the top together.

desksign
shoe

florianfesti:
OK, fixed the rectangularTriangle to work with outset edges. But as I said it might not be worth bothering with feet. Especially if the signs are typically quite small. Feet do have the benefit of things standing a bit more stable on uneven ground and ignoring bumps and ridges they are standing over. Not sure of a sign on a desk really needs that.

J-Waal:
I added the option for feet anyway. Had to overwrite the default settings because the edge is relatively short.  Also added the option to mirror one of the triangles.

PirateChest seems to contain some redundant or unreachable code left over from developing. Maybe some of it can be removed.

Do you want separate pull requests for all generators?

florianfesti:
> Do you want separate pull requests for all generators?

Separate pull request are not necessary but separate patches are appreciated


> PirateChest does not work for n < 3. Is there some way to output an error when this happens?

You can raise a ValueError with an error message. Those are shown to the user while all other exceptions are hidden. An easier option might just be to do `self.n = n = max(self.n, 3)`. Just put the minimal value into the description text of the option or even of the generator as a whole.

J-Waal:
Here is the picture for PirateChest.

The input condition is checked and shows an error to the user on invalid options. The last update includes this requirement in the parameter description.

Is there anything else that needs to be done?
PirateChest

florianfesti:
OK, I split the commit and added the sample images. I merged the first 3 generators into upstream master. The CompartmentBox might deserve a second look. I wonder if we can reuse some code of the TypeTray.

florianfesti:
OK, I was able to delete quite a few lines of code - including the sides. I think the lid wasn't quite right and I think I fixed it. But I might be mistaken. This definitely needs some test cuts before leaving "Unstable". Probably for different top edge and grip combinations.

florianfesti:
OK, I squashed the commits. I think all know issues are handled. May be a picture of the box wouldn't hurt but otherwise I think this is more or less ready to be merged.

florianfesti:
This has not moved in quite a while. Any chance you still get this test cut? Or should I cut a small model?

J-Waal:
It would be great if you can make a test cut. I do remember having tested the model and updated the margins, but was never able to produce the new version.

If everything is correct you can merge and close the pull request. If anything is unclear I may need to add some extra documentation in the code.

florianfesti:
OK, finally had a look, did a test cut, fixed a few minor issues, made pictures and pushed as 9acff7f9b357b10bf62342ba3a7b98ef6dbfecb0

Sorry, this too way too long. I have the sneaking suspicion that there may be a few more bugs. But I am not testcutting more options. I at least had a quick look over the generated drawings.

PR #551, opened 2023-04-09
Fixed flanged front wall top edge spacing in rack19box

viniciusvillas:
You just have to render any rack19box to see the bug in the rendering of the front panel. This code fixes it. Cheers!


florianfesti:
Thanks for the fix!

PR #566, opened 2023-06-21
backslashes not showing in docs

deepinthewoods:
backslashes need to be escaped for the windows paths


florianfesti:
Thanks for the fixes!

PR #571, opened 2023-07-16
fr translation: fix typos

mmuman:
Some typos I noticed.


florianfesti:
Thanks for the patch!

PR #587, opened 2023-09-29
Correct format specifier

gauthier12:
Hi, 
I had to patch a format specifier in order to get boxes works locally on my computer.
Default precision specifier is not `{var:.f}` but `{var:f}` (at least on my PC with python 3.11.5)
I'm not sure if it universal but here is the patch
Cheers


florianfesti:
Thanks for the patch! I already merged #590 before seeing yours which is exactly the same change.

PR #592, opened 2023-10-07
Add new KeyHolder generator

tljuniper:
This adds a new generator with a row of hooks for small items such as keys, cables or small bags.

The number of hooks, their size and distance to each other as well as the size of the back plate are customisable. I did some test cuts and ended up with a rounded shape for the top of the hooks to make sliding items on the hooks easier.

There is also the option to add mounting holes although some playing around with the settings is needed so that they don't overlap with the hook fingerholes.

From my perspective, this is pretty much done. Please let me know if I need to make any changes or if I've missed anything. I would also love some feedback on the generator name, as well as the parameters and their defaults, since I only know of my use case.

In any case: Thanks for the awesome project! This was a lot of fun to work with and I've used boxes.py for creating quite a few boxes by now :)


tljuniper:
I've just rebased to fix a merge conflict in `samples.sha256`.

@ffesti Is there anything I can do to get this merged? Or would you like me to close this instead?

florianfesti:
Sorry, this one fell through the cracks.

Thanks for your work!

PR #593, opened 2023-10-15
CoasterHolder generator

Lavisx:
This is my first work in this project. I hope that it is a correct style of code.


florianfesti:
While this looks really nice it is very similar to the [NotesHolder](https://festi.info/boxes.py/NotesHolder). I wonder if this could just be an style option for that. Not necessarily something you need to do yourself. What do you think?

Lavisx:
@florianfesti Yes, it's similar, but unfortunately, it didn't suit me. In my design, the walls engages into the base, while in NotesHolder, the base engages with the walls. If there was a version in NotesHolder that addressed my needs and settings, I would certainly welcome it. However, I wouldn't dare attempt to combine both on my own.

florianfesti:
OK, I have moved a few features to the NotesHolder generator. Guess there is still more that could be done. But as there hasn't been happening much for the last month I am closing this now.

PR #606, opened 2023-11-26
Generator for Lego Brick Sorter

fidoriel:
Hey,

this is my approach to a generator hat produces files for a Lego Brick Sorter. It generates 5 Sieves and a Bottom which catches the small leftovers. All 5 layers can be nested, so that the smallest fits into the second smallest and so on. The height of the levels also adapt so the stacked system will be as high as the outer most level. They also are stackable so that if in use they do not slip around.

I did not have time to do a testcut neither to do a test print on paper. Images and fixes will follow maybe not that soon but they will ;)


florianfesti:
You should use a "S" type edges for the top to properly stack the sieves.

Indenting the `description` text like this will render it as pre-formatted text in mono font (it's Markdown). Yes, the Python editor like the indentation but here it just messes things up.

fidoriel:
The "S" for the edges on the top may be correct if I want the sieves to be stackable, but I also want them to be nestable. Therefor going to down to the bottom, each sieve should have smaller dimensions, so they stack onto each other but also stack inside each other for storing. Maybe just using "h" with a increased distance to the bottom will get the job done?

As an example:
|         |
|         |
|---------|
||       ||
 |       |
 |-------|

Then they should stack away like:
||       ||
||       ||
||-------||
|---------|
|         |

So I need just the absolute dimensions from the most outer sieve to store all of them nested inside each other. I think using "S" for the top edges will not do the job, but using it for the bottom edges will not do it either. 



Offtopic:
How can I select "S" within "ABox" for the top corners? This seems hardcoded to me. I just have a dropdown for the bottom edges.

Rotzbua (changes requested):
Please remove the protected name "lego". The company's trademark lawyers are very aggressive and could cause unnecessary trouble.

boxes/generators/brick_sorter.py
```diff
@@ -0,0 +1,180 @@
...
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
Rotzbua, line 15:
```suggestion
from typing import Tuple

```

boxes/generators/brick_sorter.py
```diff
@@ -0,0 +1,180 @@
...
+    def _calc_hole_count(self, inner_mm_after_nesting: float) -> int:
+        return int(
+            (inner_mm_after_nesting - self._sieve_grid_thickness)
+            / (self._level_hole_size + self._sieve_grid_thickness)
+        )
+
+    def _calc_grid_size_width_offset(
+        self, inner_mm_after_nesting: float
+    ) -> (float, float, int):
```
Rotzbua, line 103:
```suggestion
    ) -> Tuple[int, float]:
```

Rotzbua (changes requested):
Just a tiny typing error

florianfesti:
Yeah, then the "S" type edge doesn't make sense.

You can select the "S" type top_edge in `ABox` by switching over to `UniversalBox`. `ABox` is a simplified box without too many variants and options.

florianfesti:
Hmm, this is still a Draft. Is this on purpose? Otherwise we might be able to get this merged. Although a pictures might be nice.

fidoriel:
It is not forgotten. We will get a Laser Cutter at my university, but it is currently stuck in a logistics hell. ASAP this thing is usable, I will test cut this PR and add images. If you like, merge it. I have never tested it in any form.

fidoriel:
Still not forgotten, still need to get a Laser Machine to try it out, my university is waiting for it.

littleroblet:
@fidoriel I saw your pull request when I was taking photos for some boxes without images. I have been testing things with my machines and doing some tweaks/maintenance so needing things to cut. I was intrigued by your brick sorted so I made a smaller version. I hope you do not mind me downloading your generator to try. 

I was a bit perplexed when I put it together I was a little tired and I came back to your post and then it made sense that I stacked it upside down. Once I figured that out it looks much better. I did drop some bricks through and it worked pretty well.

I love this box and the idea, it is a super cool thing. I am going to have to get a big box of bricks to play with this more. Any excuse to buy Lego? 

Here are some images I took. If you would like to use these images as the sample while you wait to get laser cutter time. You can use the one that is attached or let me know and I can take a better one in my lightbox. I can put some Lego in it for some action shots. Well, I am not a great photographer but I can try. 

Boxes laid out
IMG20240302185036
Staked the right way
IMG20240304105545
Stacked for storage
!IMG20240302185108

The ones I took when I was stacking for storage rather than use. 
IMG20240302185046
IMG20240302185053
IMG20240302184949

I have some feedback but this is all nice to have(The current box is great already). Although these things could be done later by you or someone else making improvements:
1. They stack a little tight, I think this is good but the biggest box was just a smidge tight. This could be the ply I am using but maybe a setting for extra wiggle would be nice to have. This is more of an issue when stacking upside down for storage
2. Option to generate all sieves at once. 
3. A setting to increase the overhang height when stacking would be good. It is a very small amount of bottom edge to stack and it would be easy to knock a layer off. 
4. An option to have all the boxes the same size and stack with the stackable box style rather than the overhang. (this would be good for if someone only wanted two sizes)
5. The X, Y, and H settings mentioned they are inner size but this is the outer size. Normally there is a checkbox to make this the outer. You mention it is the outer in the text description but just a little confusing having the text reference inner when it is not and there is no option for it to inner (I think this might be just how boxes work, so may not be easy to change)
6. The description uses H1 to start it and it is a bit dominant, maybe drop this to H4. This is the size I used on the SideDoorHousing box description when I updated that (whoever designed that box is awesome).
7. The description text overruns the page and does not wrap which makes the page a little messy. I think this is what Florian mentioned above about the indenting. 

Thank you for designing this cool generator. It is so neat. Cannot wait for it to be live.

fidoriel:
Thank you for testing and the great feedback. Thanks for hinting ad the wiggle, I quickly added this. Thought about doing it, but it slipped through. Also formatted the text and changed the heading.

Generating everything at once is out of scope for me, sorry.

Also, it does say "Inner Width" but most outer width is correct. This is reasoned in the base class. @florianfesti how to change that most efficient?

100mm outer box fits exactly with dimension in the svg.
Bildschirmfoto 2024-03-04 um 07 40 16

@florianfesti the lip for stacking is also given by the base class. My idea is to go back to the "s" edge, this will help. Is there a better way to increase the lip? If you have a better idea, I will revert the commit.

littleroblet:
Oh you are too quick getting some stuff updated. It is such a great idea. Glad to see you are still working on it. 

The idea about generating all at once is very much a nice to have. You can't do everything. 😀

florianfesti:
See https://github.com/florianfesti/boxes/blob/master/boxes/generators/abox.py#L40 for an example on how to adjust the sizes. `adjustSize()` default to adjusting by one thickness on each side. But you can pass None, or an edge object or character to use another value.

florianfesti:
OK, this looks pretty much finished. If you can press the "Ready for review" button above we can get this merged. I'd add a few of @littleroblet 's pictures to the commit. Although if @littleroblet could provide one or two with a bit more neutral background that'd be even better. But we can do with what we got so far.

fidoriel:
I am planning to cut this in the next two weeks and will provide images out of a white photobox. The lasercutter is now in out shop, sorry for the delay :)

fidoriel:
Now I was able to cut, take photos and fix some bugs. 
For the images: Do I need thumb images for images used in the readme? Why does it not auto add images?

littleroblet:
@florianfesti sorry I have been away and busy with work and not had much time to spend on this. Looks like @fidoriel has taken some images so I wont bother. 

@fidoriel there is some documentation on how to provide images https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#provide-photos-for-generators Florian and I updated this documentation to be a bit clearer recently. Let me know if you have any issues. If you are really stuck attach the images here so @florianfesti can add them when he merges.

fidoriel:
Thx. Added images and tested and fixed everything.

fidoriel:
@florianfesti Test for Shadowbox fails. It seems that the box does not have a example svg. https://github.com/florianfesti/boxes/commit/0f30cbd4a4d6eb52ad4381ec1a413569bcb31a44 did not add this. Same issue on my other PR: https://github.com/florianfesti/boxes/pull/645

florianfesti:
Don't worry about that. Those test are pretty new. I have not 100% decided if I really want them to fail for new generators. I'll have a look at that on the week end.

florianfesti:
OK, finally...

florianfesti:
Squashed and merged. Thanks for your work!

PR #613, opened 2024-01-06
Add Blender Integration for Boxes.py

Franckapik:
This is the Pull Request for integration of Boxes.py as a Blender add-on.
Documentation updated and installation instructions on the website of blenderCAM used for CAM on Blender3D : https://blendercam.com/blender-boxes/


florianfesti:
Uff, this is totally not what I meant when I suggested to commit it in the main repository. For this we need to find a way to not duplicating the whole project. I though we can do this in a similar way as the Inkscape plugin that mainly uses the Boxes.py library and generators as they are and just generates some interface files and has some wrapper to translate between both projects.

Now your question about needing pull request makes a lot more sense. But the answer is: You need a solution where this is not an issue because your changes are generic and updates will not needed regularly.

Franckapik:
Hi Florian,
I understand your last message. I was suggesting this PR to have a look on the blender integration. I have duplicate the main repo because Blender need a zip file with the scripts and depedencies inside unless we can install them with pip. 

I can try to work on add Boxes package to Pypi and then work on a installer for both blender and Inskape. I quicly read this tutorial and it seems that i can prepare the boxes folder for PyPi. But i probably need permissions on github repo :/ I understand  that you probably don't have the time to work on this .

https://medium.com/@joel.barmettler/how-to-upload-your-python-package-to-pypi-65edc5fe9c56

ps : i am a french developper so we can take time to discuss about it in another way ;)

Thank you !!

PR #625, opened 2024-02-10
Add settings for bed bolts

chrysn:
This adds user configurable Settings for bed bolts. It's a *bit* odd because so far all previous pluggable Settings were for edges (and bed bolts are not really edges, but more like a tool that fingers can use), but it still works well.

To have an example, the closedbox (which already had bedbolts in the source but they were commented out) gains a configurable option to enable them.

There is one more change that'd make sense in my opinion, but it'll prefer to discuss that a bit first: I think that edge.Bolts should grow a second argument "layers", which defaults to 1 and is ignored on the drilling side. The total length is reduced by thickness * layers. This way, we can have the user configurable length parameter the length of the screw (plus tolerance), and depending on the number of sheets (with round holes) this will be screwed through, the length will be reduced accordingly. After all, as a user configuring this, I'll know the length of my screws, and I'll configure the thickness, but I'd use the same screws everywhere, and if there is in some place a constellation of two layers (which may happen in the box I'm doing this for), that should be accounted for by the box script.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -389,7 +389,6 @@ class Boxes:
         if self.ctx is not None:
             return
 
-        self.bedBoltSettings = (3, 5.5, 2, 20, 15)  # d, d_nut, h_nut, l, l1
         self.surface, self.ctx = self.formats.getSurface(self.format, self.output)
 
         if self.format == 'svg_Ponoko':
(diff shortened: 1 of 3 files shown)


chrysn:
I completely missed in my original check that there is also https://github.com/florianfesti/boxes/pull/468 that does something similar.

One item probably worth pulling from there is the drop-down list of standardized bolts. But I'll wait for an initial round of feedback at least on directions before proceeding here.

florianfesti:
Closing this PR as it needs updating. We still have https://github.com/florianfesti/boxes/issues/36 open.

PR #633, opened 2024-02-27
Updated the contributing documentation

littleroblet:
Updated the contributing documentation including (updated the ImageMagick wording to reference installing Legacy utilities for Convert and added running the images through tinypng


littleroblet:
@florianfesti I updated the documentation to add tinypng to it but looks like you did too. Let me fix that conflict and just add the small update to the ImageMagick

PR #642, opened 2024-03-04
Added feature: grippy corners

ojensen5115:
In a generator I am working on, it is helpful to create circles with grip on them. More generally, it seems useful to be able to create curves or corners with the same kind of corrugation as `self.grip()`.

To DRY as much as possible, I've added a new parameter to `self.grip(...)` to take a number of degrees over which it will arc (defaulting to 0), and then creating `self.gripCorner(...)` with an API similar to `self.corner(...)`

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1017,18 +1017,30 @@ class Boxes:
 
         self.move(tw, th, move)
 
-    def grip(self, length, depth):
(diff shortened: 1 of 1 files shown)


ojensen5115:
Ah, I see we already have "waivyKnob" which is what I actually needed this feature for. I'm happy to close this PR as not needed unless anyone else has an actual use case for a grippy curve.

PR #643, opened 2024-03-04
New generator: FlexBook

ojensen5115:
A box with flex that is styled after a hardcover book. Photos included.

image


florianfesti:
Very nice work! Thanks for the generator!

ojensen5115:
@florianfesti When I construct a piece from edges and corners and put it next to a `rectangularWall`, it ends up slightly lower and to the right -- it's like there's some kind of special margin built in to the pieces that doesn't get applied when constructing pieces out of edges. Moreover, I've found that even inside the piece, things can be off by the width of the burn correction.

Here is a minimal example:
```
def pinHoles(self, width, height):
    t = self.thickness
    self.rectangularHole(width/2, 1.5*t, t, t),
    self.rectangularHole(width/2, height-1.5*t, t, t)

x = 30
y = 20

self.rectangularWall(x, y, move="right", callback=[lambda: self.pinHoles(x, y)],)

self.pinHoles(x, y)
self.edges["e"](x)
self.corner(90)
self.edges["e"](y)
self.corner(90)
self.edges["e"](x)
self.corner(90)
self.edges["e"](y)
self.corner(90)
```

In the output, the two pieces have a small vertical offset:

image

After aligning the pieces to overlap perfectly, the pin holes are vertically off by the burn correction (in my case 0.12 mm):

image

Since I don't understand what's happening here, I've generally avoided mixing the two styles of constructing pieces.

florianfesti:
There are two different things at play:

`Boxes.move` - which is used by all parts (see https://florianfesti.github.io/boxes/html/api_parts.html) - adds a bit of space round the pieces to they do not touch each other. The value used is `Boxes.spacing`. This is the reason why the rectaqngles are not alligned the same way. Using `Boxes.move` would fix that.

The second thing is the burn correction. See https://florianfesti.github.io/boxes/html/api_burn.html

Basically the bottom line is one `Boxes.burn` below the nominal area of the piece. As such the holes in the middle need to be one `.burn` higher. `.cc()` that handles callbacks does that semi automatically by having the default offset for y as `None` which is translated to `Boxes.burn`

Because of these issues it is better to use callbacks as this separates the caring for those offsets from the stuff being drawn onto the piece.

PR #644, opened 2024-03-05
Make better use of part names

chrysn:
So far, part names are only used to place red text labels.

This change stores the names in the parts (the constructor argument was previously ignored; an extra argument is added to new_part to account for its typical usage pattern where drawing happens on the last set part, and a new part is created at the end of drawing).

A second commit uses the data that is now stored to make names accessible in browsers' tooltips. Future changes enabled by this would be making parts more easily accessible in DXF as well, eg. to extract them for further processing into 3D models.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1217,7 +1217,7 @@ class Boxes:
                     self.moveTo(x, 0)
                     self.ctx.scale(-1, 1)
                 self.moveTo(self.spacing / 2.0, self.spacing / 2.0)
-        self.ctx.new_part()
+        self.ctx.new_part(previous_part_name=label)
 
         return dontdraw
 
--- a/boxes/drawing.py
+++ b/boxes/drawing.py
@@ -536,6 +536,8 @@ Creation date: {date}
                 continue
             g = ET.SubElement(svg, "g", id=f"p-{i}",
                               style="fill:none;stroke-linecap:round;stroke-linejoin:round;")
+            if part.name:
+                ET.SubElement(g, "title").text = part.name
             g.text = "\n  "
             g.tail = "\n"
             for j, path in enumerate(part.pathes):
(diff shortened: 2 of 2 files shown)


florianfesti:
I very much like what this is aiming at. I am not that much a fan of the implementation. Let's just make sure we pass the `label` to the first call of `move()`, too. `add_part` might need a bit of tweaking to make sure we don't keep the old empty part with the old label. But that should just be a single line.

chrysn:
Oh I fully agree that the retronaming is not pretty. I've tried to make all changes such that no generators would be changed, given that boxes.py can also be used as a library and I don't know what the API guarantees there are.

If I start an alternative version of this where there is a mandatory new_part called at the start of a part rather than at the end (which also makes later steps easier), is there a test suite somewhere that I can run that will execute all generators at least with default configuration? (That'd be helpful because I could make it scream the right way until every move start matches its move end). Ideally, is there something that I can run that asserts that the box outputs did not change?

chrysn:
And because I'll have to deal with the temptation: Is there a rationale for why the move blocks are not implemented using context managers? (The refactoring necessary to do the labels in the first part will note quite warrant that kind of change, but the temptation will be there).

florianfesti:
I think the main reason is that context managers did not exist in Python when the code was written. Also passing width and height is super annoying. This was necessary in the past as Boxes.py used libcairo. As we now use our own backend, we should be able to calculate the space needed by the pieces and place them correctly accordingly.

florianfesti:
Wait a second. `Part` now has both `.extents` and `.transform`. This might be easier than I thought. Give me a minute...

chrysn:
Would you care to give me some more pointers as to what you'd like to have simplified? (I can also start refactoring the label assignments based on giving the part name beforehand, but it seems you have more concrete ideas.)

florianfesti:
OK; this minute stretches on...

I do have a branch that detects part sizes and uses that for move(). It also sets the group names. But I currently don't quite have the time and attention to get this finished soon. I still think it super seeds this PR. SO I am closing this here as keeping it open indefinitely isn't helping anyone. Just know this has pushed the project into a new direction even if the change is getting implemented rather slowly.

So thanks for your effort!

PR #647, opened 2024-03-11
New generator: CoinBankSafe

ojensen5115:
A piggy-bank designed to look like a bank safe.

image

Modeled after https://3axis.co/laser-cut-wooden-safe-box-6mm-mdf-cdr-file/075yqnko/


florianfesti:
This looks really neat already. I'd suggest adding the number from 1 to 8 to the dials and add a second square hole to the number dials that is 45° rotated. That way they can be added in each position  during assembly.
You might want to add a mark above the dials to indicate where to read the number.

When it comes to the code there are a few things that can be achieved easier when using more advanced features of the library. But that is something we can also look at later on. Using `parts.disc` and `parts.waivyKnob` with a `callback` and `move` parameter is the first thing I stumbled across. This cut-off disc is probably something we want in parts, too. May be even as a parameter for `parts.disc` itself.

ojensen5115:
Thanks! I will update the designs to include etching the numbers 1-8. I think I would either cut out twice as many cut-off discs, or making a parameter to ask how many of the discs should be cut with a 45-degree square hole, because I'd worry that especially at 3mm having a star shaped hole would not be very strong. Though 🤔 perhaps with the glue it wouldn't matter, I'll need to test it out.

I looked briefly into using Discs, but I did not see an easy way for it to give me a square hole. I think extending `parts.disc` would require adding a flag for square holes, and also adding a parameter for how much of the disc to chop off, which felt like a significant enough change that I didn't make it. I'm happy to do so, though, if you prefer.

florianfesti:
`parts.disc(d=d, callback=lambda:self.reactangularHole(0, 0, t, t))` should do the trick.

florianfesti:
For larger things - like adding numbers - you can use a function or method instead of a lambda expression.

ojensen5115:
I haven't really explored the callback functionality yet, but I will do so and update this PR when I have some time this weekend or next week.

Thanks for the help / pointers!

ojensen5115:
I've added the numbers and made an extra hole at 45 degrees on the D-discs. I still need to experiment with parts, in particular to improve selection grouping in the end result.

ojensen5115:
I have made the following additional changes:
* rotated the door handle to match the door grain
* updated the locking pins with barbs, which means you don't need to glue the locking disc (making it easy to change the code)

florianfesti:
OK; I move the code to use callbacks. This allows using the standard pieces at more locations. I was able to get rid of circleSquareHole completely, even if the callbacks are a bit hacky now.

florianfesti:
Overall this looks pretty good now. Anything else you want to add? Otherwise I'd squash the patches and merge it into master.

ojensen5115:
Oh that's a ton better, thank you.
I just have a minor improvement to the dial numbers and better photos, i'll get those committed shortly.

florianfesti:
OK, ran the picture through tinypng to save a few bytes. Time to merge!

Thanks for your work!

florianfesti:
At some point we need to add a one dial, 3 number lock like a proper safe.

PR #654, opened 2024-03-28
New generator: Shadowbox

ojensen5115:
You can find some great vectors online to cut out layers for a shadowbox (also called a lightbox), such as this one here:
https://3axis.co/laser-cut-my-neighbor-totoro-3d-lightbox-lamp-cdr-file/eoxldrxo/

But once you've cut out the paper sheets, you need some kind of frame to space and mount them into. That's what this is.


florianfesti:
Wow, this looks gorgeous! There is still a left over print in line  103.

Otherwise this looks really good!

ojensen5115:
Whoops, thanks for catching the print statement. Fixed!

florianfesti:
OK, squashed the patches and added a bit nicer commit message. Thanks for the patch and these beautiful pictures!

PR #672, opened 2024-04-25
Shadowbox improvements

ojensen5115:
**Commit 1:**
I made an error with the angles when refactoring, rendering the angle complement calculation ineffective, making them wrong when the angle is not 45 degrees.

**Commit 2:**
I have also added parts to glue the LED strip onto.

**Not yet committed:**
The look of the shadowbox is *greatly* improved by adding a frame around it with nicer wood:
PXL_20240330_234543198(1)
PXL_20240330_233651879(1)

However, this will tend to be of a different thickness than the main construction (e.g. 3mm vs 6mm). I couldn't figure out how to have two different widths apply to finger joints in a single generation, so I don't know how best to do this. To do it myself, I just added them to this generator, ran it twice with different parameters, and manually combined the resulting files as needed.

I suppose I could make a separate "picture frame" generator, but this would require entering a bunch of parameters that are calculated values here.

How would you advise doing this?

### DIFF
--- a/boxes/generators/shadowbox.py
+++ b/boxes/generators/shadowbox.py
@@ -107,3 +107,9 @@ See the diagram below for dimensions.
         self.rectangularWall(x, height, f"ef{top_edge}f", move="up")
         self.rectangularWall(y, height, f"eF{top_edge}F", move="up")
         self.rectangularWall(y, height, f"eF{top_edge}F", move="up")
+
+        # led strip holder
+        self.rectangularWall(x - 2*t, 10, "efef", move="up")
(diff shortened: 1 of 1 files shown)


ojensen5115:
With some further thought, I think the least silly way to accomplish the exterior frame is to make a separate generator. That way, it's also useful if people want to literally make picture frames or similar without having to generate a shadowbox and delete the shadowbox part :) It's not too onerous to measure the resulting sizes to input into the frame generator parameters.

...or I suppose I could even update this generator to add a text-string label somewhere that gives you the parameters to use in the frame generator if you want to 🤔

PR #680, opened 2024-05-28
Add photo frame generator

marauder37:
3-layer photo frame. 
    
    Selected excellent features:
    
    * easy to change the photo after glue-up, without disassembling the frame
    * calculates the ideal matting size for your photo based on ancient Greek mathematics
    * can make the frame in one piece or split into 4 pieces to save material
    
    Features available in the mysterious future:
    
    * add a hole for hanging the frame on the wall
    * rounded corners (works now on selected layers)
    * calculate the frame size based on the piece of glass/acrylic you already have
    * determine the width of the internal area based on the thickness of the matting and glass

sramanora
sea
klein
frames


florianfesti:
Sorry, for the bit late response. I am currently polishing this up to pass the test suite. I added one of the pictures as a sample image and reformatted the description.

I am wondering if we really need this many parameters or if e.g. the frame width can be the same for width and height.

I also wonder if having both split and unsplit parts (or neither) is just confusing and we should only offer either or.

marauder37:
I just picked this up again after a while. My supply of pre-cut acrylic arrived, so I'm cutting more test pieces out of plywood and seeing which parameters I miss if they're taken away. Since the UI params are completely decoupled from the dimensions it's easy to change this stuff around. 

Frame width should be fine as one number. I have made it one parameter in my local version.

Split and unsplit parts, I'm still not sure about. You're not "supposed" to want split and unsplit at the same time. In plywood you'd probably always split the middle layer and then either split the top if you're cheap like me or cut it in one piece if only the best will do. So my current idea is just one param for "split front", default on.

But I looked at the frame in the sample image again, which was a test piece made out of cheap thin corrugated cardboard. It's actually a 5-layer construction with both split and unsplit parts for top and middle. This is because it's made out of thin material with a shockingly wide kerf, so it gets height from the split layers and stability from the unsplit layers. Maybe nobody else will ever need to do this.

What I notice most now that the acrylic has arrived is that I should hurry up and finish the features on the to-do list. Otherwise I have to back-calculate the width of the matting to fit the pre-cut acrylic, and figure out how to hang the frame. For that I will look at MountingSettings and see if it applies here.

IMG_1731

A small thing that I keep noticing is, the descriptions for `x` and `y` parameters are "inner width in mm" etc. If I was writing the help text, it would be "width of photo in mm, excluding border". I haven't looked at how to... reach in and change it. Changing help text for default options might affect l10n?

For matting and framing, users need to know exactly what they're supposed to measure. Otherwise they need to guess whether the generator will factor in all the allowances and overlaps needed for proper framing, or if they should add fudge. Or worse, users who haven't matted artwork before will learn about matting the "hard way" when their photo pokes through the mat. I'm trying to have a "don't guess, just measure precisely and tell the generator what you've got" UX.

florianfesti:
Wrt the `x` and `y` parameters: Just drop them from self.buildArgParser() and add them manually with self.argparser.add_argument(). The description is semi automatically added to the translation files (when the project is build and the changes are commited) and can then be translated. There is nothing you need to do from the programming side of things for text in the arguments.

marauder37:
Thanks! I think I've now pushed this feature as far as I want to push it, unless there are bugs or it needs rework

florianfesti:
Thanks a lot for all your work! Even this is "just" a picture frame it has seen a lot more love and attention to detail than any of my generators.

PR #690, opened 2024-07-02
Add USB-C support to MakitaPowerSupply

foosinn:
This adds optional USB-C and no secondary power supply support to the MakitaPowerSupply


florianfesti:
Added explicit default value to  `secondary_power` option to keep current behavior in the CLI and keep the pre-commit checks passing.

PR #699, opened 2024-08-20
Add new stackable bin generator

pkochubey:
A regular box that can be stacked on top of each other.


pkochubey:
@florianfesti Thx, it's very good suggests. I fix code and image, now tray stable!

pkochubey:
But now I look at it and understand that the “limiters” can be glued on and it will look even better inside.

florianfesti:
For the grip and label piece I was thinking about a cross section like this:

```
|
|
|                       /|
|______________________/
```
To be fair, this is pretty difficult to see on the photo. This way you can grip the bin from underneath and pull it out. Ofc this requires the sides to extend to support this front piece.

boxes/generators/stackablebin.py
```diff
@@ -0,0 +1,87 @@
...
+        self.addSettingsArgs(edges.FingerJointSettings, surroundingspaces=0.5)
+        self.argparser.add_argument(
+            "--front", action="store", type=float, default=0.4,
+            help="fraction of bin height covered with slope")
+
+    def render(self):
+        if self.outside:
+            self.x = self.adjustSize(self.x)
+            self.y = self.adjustSize(self.y)
+            self.h = self.adjustSize(self.h, e2=False)
```
florianfesti, line 58: This should be adjustsize(self.h, "s", "S") to adjust for the stackable edges

pkochubey: done

pkochubey:
IMG_20240826_180927
Fully tested 12x12 cm

florianfesti:
OK, I added a few more fixes. I was pretty busy last week. So it's easier to just drop the changes now and here.

After agreeing that we are happy with the result the patches should all be squashed into one.

pkochubey:
@florianfesti Looks good, thanks

florianfesti:
OK, squashed most commits. I moved to the build in bottom stabilizers after fixing them. They are a bit shorter to they work with stackable edges all around. They still should be fine here.

florianfesti:
OK; this has gone on long enough. Thanks for all your work!

Issue #563, opened 2023-05-26
Request : Dice Tower


noscollections:
To complete Dice objets, a dice tower will be great.

A simple tower like this one in the design : https://www.thingiverse.com/thing:2001685

On parameter, we could have :
The height of the tower
Section of the tower,
number of intermediat inside section with angle
size of the arrival zone of the dice.

I can make a draw if you think this a good idea

tljuniper:
With #704 merged, can we close this issue or is there anything left to do?

florianfesti:
Sure!. I forgot about that. Thanks for reminding me!

PR #704, opened 2024-09-08
Add new DiceTower generator

tljuniper:
This adds a new generator for a simple Dice Tower.

Size of the tower as well as the number and angle of the ramps inside the tower are configurable.

I've opted not to add a separate area for catching the dice. If desired, it's always possible to add a matching ABox and glue the tower down in it (or keep the two separate for easier transportation).

This implements most of the requirements in #563. @noscollections I'd love to hear your feedback on this solution! :)

image


noscollections:
That look very good. If a html test link is available to choose parameters and test lazer cut. I can test it quickly.
I would like to test because a parameter for the height of dice exit could be interesting. That depend of dice size.
I need to test with all my dices. I have D6, D8, D10, D12, D20 D30 D100 Dices. Need test for the big one.

Yes a separate extensions could be made. Square or hexagonal (like a false circle)

florianfesti:
Looks like the bottom piece is not really needed. Otherwise this looks pretty nice already.

This is a very welcome addition as I have procrastinating on this for years as I could not settle on a set of features. Just having a simple one is probably the best solution.

tljuniper:
@noscollections 

> a parameter for the height of dice exit could be interesting.

In the default tower, the size of the dice exit is around 4cm.

The size of the gap is not fixed but calculated based on the angle of the ramps and the depth of the tower: it starts at the same height as the top of the last ramp, so it's `depth * tan(angle)`. So for an angle of 30 degrees, `tan(30) = 0.58` -> the opening is a little larger than half the depth of the tower. If you don't choose a smaller angle, then the most narrow part of the tower will be at the upper ramps (3.7cm in the default tower).

> If a html test link is available

With Python installed on your local machine, you can check out the PR and start a local version of the website, see [here](https://florianfesti.github.io/boxes/html/CONTRIBUTING.html#running-the-code). I don't know of any way to have a ready-made html test link for a PR, sorry.

tljuniper:
@florianfesti 

> Looks like the bottom piece is not really needed.

Yeah, that's true. It really helps when you'd like to glue the tower down or if you'd like to add some weight in the bottom though... Since we don't have too many parameters yet, I've added a checkbox for including/removing the bottom piece. Does that seem okay?

> as I have procrastinating on this for years as I could not settle on a set of features

I may or may not have started working on this half a year ago with the plan to make it a birdhouse style tower (like in the game wingspan) and to add the option for a castle-style top edge :joy:

florianfesti:
Let's merge this quickly before we get more ideas for additional features. More features can always be added later. This thing has waited long enough for stuff that is never going to happen...

Thanks for your work!

noscollections:
Tested today. Perfect for me.
For the first I test default parameters. Perfect for my dices D6 - D8 D10 D12 D20.
But no OK for the D30. (i don't use it many time. Juste need to add few mm on the height.

Next time I try a 320mm height.

Will try the OpenBox script to do a dice receiver.
The RegularBox is also a good script for the receiver, with 6 ou 8 sides, but need to have a side open
And we also need to give the size of one side (the open one) to have the same size of the width of the tower size.

But with some paramters test We can do it .

Thank you for the great work

PR #705, opened 2024-09-08
Add new BookHolder generator

tljuniper:
This adds a new generator for a book holder.

It's a bit similar to the Display but the construction is more robust and it has the option for a front ledge to hold books open or hold items in place.

image


florianfesti:
This totally should use "h" edges to be more robust. But I just can't see how to assemble it then...

Guess it stays as is for now. Thanks for the new generator!

Issue #628, opened 2024-02-13
Sample Images missing


florianfesti:
Many "Box" generators still need a sample image. Please donate one if you create something that is still missing a picture! See [Contributing Document](https://github.com/florianfesti/boxes/blob/master/CONTRIBUTING.rst#provide-photos-for-generators) for details or just add the picture in a comment below.

The following generators are still missing a picture:

- CarbonFilter
- BottleStack
- DinRailBox
- Dispenser
- Display
- FillTest
- FlexBox5
- FlexTest2 - could be rendered image
- Gears
- Hook
- MakitaPowerSupply
- NemaMount
- Planetary
- PoleHook
- Pulley
- Rack*Box
- RectangularWall
- SevenSegmentClock
- SideDoorHousing
- SkadisBoard
- StorageRack
- Tetris
- TriangleLamp
- WaivyKnob 
- WallEdges
- WallRack

There was an old ticket (#140) for this which got too long.

D-Platoon:
It was not asked, but I have a new picture for the ConcaveKnob (haven't seen the WaivyKnob).
Maybe you can use it.
![20240222_163210](https://github.com/florianfesti/boxes/assets/159539656/10f32668-3a73-4f37-8b1a-6b20db8d2dbd)

littleroblet:
I have added some images on Pull Request https://github.com/florianfesti/boxes/pull/632

Flexbox5
SideDoorHousing
Tetris

_I just wanted to test a few different things on my laser cutter so I picked a couple without images so I could contribute the images._

revjmoney:
![20240815_024558](https://github.com/user-attachments/assets/837f8ffe-3fc0-4696-82e0-a4ba26896b3a)

Sliding Box Generator.  I needed a container for the metric tap set that amazon sent across the country loose in a plastic baggie?  thankfully they arrived intact.  Laser cutter for the win, now they have a home!

ClaneSharke:
FlexBox:

![Image](https://github.com/user-attachments/assets/12143443-be7f-4417-88a0-163cef268393)
![Image](https://github.com/user-attachments/assets/d32d8bdb-9020-43f3-913e-5e9289bd8dcf)

Cemelinio:
I think there is lots of room for improvement in the awesome original generic 7-segment clock template:
[...]

PR #708, opened 2024-09-16
Add image for Display

tljuniper:
Had a look at the Display and saw that an image was still missing (#628).

The `gen_thumbnails.sh` script reported an error that could be fixed by updating the hash for PhotoFrame, so I've added that change here as well.

```
❯ ./gen_thumbnails.sh        
File ../static/samples/PhotoFrame.jpg changed
sed: -e expression #1, char 26: unknown option to `s'
```


florianfesti:
Thanks for the picture!

PR #713, opened 2024-10-01
Hobby case generator 

JGrzybowski:
Hey, I've been working on hobby case generator (inpired by FrontierWargaming portable paint stations) 

I'm not fully fluent in python and had some issues with positioning the finger holes so I'm open to any suggestions. 

I'll add the pictures once I get the fixed shelves burned (I screwed base plate and top/bottom by thickness of the material)

I see the potential for several upgrades in the future, but want to close the v1 and learn what could be simplified with the generator as it is.


florianfesti:
Sorry, for not writing earlier. This looks very interesting. Although it is a bit more complicated than it probably needs to be. I just pushed a commit that moves the fingerHoles into callbacks (see https://florianfesti.github.io/boxes/html/api_parts.html#the-callback-parameter) for details. This gets rid of the brittle placement.

The whole segmented_edge business it technically not needed.

You can use a `type=argparseSections` for "--unitW". See https://florianfesti.github.io/boxes/html/usermanual.html#section-parameters on the benefits.

JGrzybowski:
Thanks for feedback. 
I knew that some parts were definitely "overengineered". Now that I understand the idea behind the `callback` argument in `RectangularWall` it makes code so much cleaner.
- I'll take a look on those input args.
- The `segmented edges` is more of a helper function to reduce code duplication. I'll try to simplify it, but I couldn't figure out if there's a nice Python one-liner for such things.
- I'll take a look if the design didn't get affected by the changes (I assume it did not, but "measure twice, cut once") 
- I'll regenerate examples once I do all of the above.

I've got two questions:
- Should I squash the commits before you merge it or will you do it once you let in the PR (GitHub allows squash on merge as far as I remember)?
- Can you tag this PR with hacktoberfest-accepted label once I clean up the pointed out issues?

florianfesti:
I didn't mean `segmented_edges` itself is too complicated. Having the compound edges is not needed at all. You can just use one continuous edge on the outside. The compound edges are only needed on the inside if walls cross each other and need slots.

florianfesti:
Generally squashing the PR into neat commits with good commit messages is preferred. So a new generator just one commit is fine.

How exactly is this "hacktoberfest-accepted" label supposed to work? Are we talking about a GitHub or a Git label? I am happy to apply whatever is needed.

The parameters still need a description text. The class docstring and `description` also still need updating.

I wonder if `shelvesNs` can be a simple integer.

Do you plan to add removable shelves?

JGrzybowski:
> I didn't mean `segmented_edges` itself is too complicated. Having the compound edges is not needed at all. You can just use one continuous edge on the outside. The compound edges are only needed on the inside if walls cross each other and need slots.

Oh, I see what you mean, if I narrow the finger holes area in the back plate it should be fine. I crated those segments because I was worried that there would not be enough material on the sides.  

> Generally squashing the PR into neat commits with good commit messages is preferred. So a new generator just one commit is fine.

Ok, I'll squash it once the code is fully polished.

> How exactly is this "hacktoberfest-accepted" label supposed to work? Are we talking about a GitHub or a Git label? I am happy to apply whatever is needed.

You can add the label on the panel on the right on this PR screen.

> The parameters still need a description text. The class docstring and `description` also still need updating.

Sure, I'll fill them together with parameters cleanup.

> I wonder if `shelvesNs` can be a simple integer.

Not exactly, as each column can be of different width. Then you might need different amount of different sizes. Rails probably need the same treatment.

> Do you plan to add removable shelves?
Haven't thought of removable shelves, but rather some kind of drawers that can slide on rails or shelves themselves. But that might be an option for v2

florianfesti:
Yeah, you are not using a callback there. Move the fingerHoles into a separate function and pass this

JGrzybowski:
I've pushed the changes, but still you can see some misalignments in debug mode.

JGrzybowski:
And sorry about no squashing, but I did merge master in a couple of times during the development and it prevents me from nicely squashing things in IDE.
But I'm sure there's an option to do that on Pull Request merge (squash and merge)

JGrzybowski:
Is there anything blocking this PR from being squashed & merged?

JGrzybowski:
@ffesti, is there anything preventing this PR from being squashed and merged?

florianfesti:
Sorry, I forgot it was me who was supposed to squash this.

OK, ran the sample picture through tinypng to save space and bandwidth, fixed the formatting of the description and squashed all the patches.

florianfesti:
I'll update the example svg next...

florianfesti:
Thanks for your work and your patience!

JGrzybowski:
Thanks for help with squashing!
Can you mark this PR with `hacktoberfest-accepted` label?

florianfesti:
Like this?

JGrzybowski:
Exactly like this, thank you!

PR #722, opened 2024-11-12
Remove double-drawing of bottom polygon

cmfcmf:


florianfesti:
Yes, of course!

Thanks for the patch!

PR #726, opened 2024-12-04
feat: Gridfinite-base add pads

clemens-tolboom:
Let's not waste material by generate the pads together with the openings

### DIFF
--- a/boxes/generators/gridfinitybase.py
+++ b/boxes/generators/gridfinitybase.py
@@ -41,6 +41,7 @@ class GridfinityBase(Boxes):
             help="edge type for bottom edge")
         self.argparser.add_argument("--pitch", type=int, default=42, help="The Gridfinity pitch, in mm.  Should always be 42.")
         self.argparser.add_argument("--opening", type=int, default=38, help="The cutout for each grid opening.  Typical is 38.")
+        self.argparser.add_argument("--cut_pads", type=boolarg, default=False, help="cut pads to be used for gridinity boxes from the grid openings")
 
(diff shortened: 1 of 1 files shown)


florianfesti:
Generally I like this change. I wonder if we can make it a bit more in line wit the other gridfinity generators.

* Have the locations of the feet engraved on the bottom
* May be just make the feet 2 smaller as the holes

I also wonder if we want offer a mode with less feet as there can be a lot of them. The normal gridfinity boxes only have 4 but they do not need to put their weight on the feet but are also supported by the grid. May be offer 4, 6, 8, 9, 16?

florianfesti:
Also don't worry  about the CI. It's annoying but I can take care of it as soon as we are ready to merge the PR.

florianfesti:
I also just realized that the feet wont work with anything but the "F" bottom_edge.

clemens-tolboom:
> Have the locations of the feet engraved on the bottom

We do not need this as the pads/feet are used for the other GridFinity generator.

> May be just make the feet 2 smaller as the holes

That makes sense

> I also just realized that the feet wont work with anything but the "F" bottom_edge.

Not sure what you mean :-(

florianfesti:
Ah, I thought you wanted to add the feet to the base itself. Sure, this makes a lot more sense.

florianfesti:
OK, turned the parameter into a checkbox and made it "off" by default, so the output is unchanged and passes the CI (hopefully).

clemens-tolboom:
Thanks :-)

PR #729, opened 2024-12-08
Fix incorrect slot size on WallSlottedHolder

AlexCPU:
The left and right hand sides on the WallSlottedHolder are incorrect in width (making a 1-slot version doubly wrong in width).

I believe this PR fixes it.

Before output (on default settings).  Note the unevenness on the left and right slots:
image

Output after this PR (on default settings):
image

Edit: Any thoughts on enabling Discussions on the project?  I've got some ideas I'd like to propose, but they're not really issues.


florianfesti:
Unfortunately I don't remember the details of this generator. So I guess this is correct. Thanks for the patch!

PR #732, opened 2024-12-11
Add WallStackableBin generator and update .gitignore

AlexCPU:
Basically re-written the existing StackableBin so that it's now wall mountable.  Choose wall type "plain" to get something _close_ to the original.  Added features to choose the angle of the front sloping part and added an optional (default) vertical label section, which I think gives a more traditional stacking bin appearance.

I found trying to reverse engineer the existing generators fairly tricky, so I've been very comment heavy, but I'm aware the rest of the library is basically uncommented.

I haven't deleted the existing StackableBin as it didn't feel like my decision to make.  There's potentially a number of generators that could be upgraded with Wall edges, but that will likely balloon the Wall category, and I'm not sure if it's apparent that you can generate them without wall fixings though.

Finally I was surprised when I generated the example SVG that it wasn't listed in my unstaged files, and whilst the instructions say `git add -f examples/*.svg` I added an exclusion to the .gitignore for SVG files in the examples folder.


florianfesti:
Thanks a lot! Very nice work. Te improvements here could probably be back ported to the original StackableBin. But that's more or less unrelated to this PR.

To make the bin more sturdy one could optionally use `h` edges on the sides to the front pieces are more secure. But that makes the geometry a bit more tricky.

PR #734, opened 2024-12-11
Add WallHopper generator

AlexCPU:
I'm adding a generator for a hopper/dispenser.

I need some help on this however.  The code is a mess with two copies of everything as I've got two possible solutions:

The "left" side is created with a polygonWall, which feels like the elegant solution to me, but it doesn't compensate for the margin of the 'h' edge, and there's something wrong with the finger locations at the front/top of the hopper area.  This part does not fit with the rest of the parts.

The "right" side is created with a rectangularWall, and a modified edge.  It _works_, but to get the correct shape I had to make the custom edge go back on itself, so there's an extra unwanted line in unused space.  Cut two of these parts to get a working hopper.

If you've got any tips on how to fix either of these methods, or are prepared to accept the one with the excess line, I'll get some changes made.


florianfesti:
Just draw the margin of the "h" edge yourself. `self.edges["h"].startwidth()` gets you the outset you need depending on the current edge settings. Add a 90° corner at the right side, the startwidth() `e` edge, -a corner. At the end just add a 0° corner and a startwith() `e` edge before the final 90° corner.

florianfesti:
Alternatively you can use the startwidth as radius of the `90-a` and `a` corners and switch to an `h` edge instead of the two `f` edges at the front. Need to compensate for that somewhere else, though. Best place is probably the calculation of `b` and `tl`, but there are a few other options.

florianfesti:
Code speaks louder than words.

AlexCPU:
Thanks, I think I was about half an hour behind you with a commit!  I'll tidy the rest up and get new pictures/examples sorted later today

AlexCPU:
I actually quite like the `h` for that piece too, as then the aesthetic matches all the way round.  I've fixed all the files, and added a couple of bits where the material thickness was missing, which was throwing the hole alignment off.

I'd consider this _done_.  I'll add photos in a couple of hours when the glue has dried, and mark the PR ready to go.

florianfesti:
Thanks a lot for your work. This is a really nice generator!

Additions could be a lid for the top or even a hinged lid (e.g. made from acrylic) for the troff. Those would ofc be optional but would prevent dust getting into the content.

AlexCPU:
Thanks, I'll raise an issue for adding hinged lid options, that either I or someone else can add in the future.  I did try it yesterday by just swapping in the appropriate `j` edges, it wasn't quite as easy as that though, with overlapping cut lines, so didn't make my cut for MVP.
image

PR #737, opened 2024-12-13
Add generator for ZBeam

jasondunsmore:
I needed this to make a channel to hide a cable up against a wall corner.


ffesti:
3 `f` edges and only one `F` edge are unlikely to all fit together.

This also obviously is an U beam...

jasondunsmore:
> 3 `f` edges and only one `F` edge are unlikely to all fit together.

Should be fixed.
 
> This also obviously is an U beam...

I added a note about this. Also added an option to make a flanged U beam.

florianfesti:
Looks good. I was about to just hit merge. I wonder if you can add a sample picture. If you don't want to jump all the hoops listed in https://github.com/florianfesti/boxes/blob/master/CONTRIBUTING.rst#provide-photos-for-generators I am happy to deal with that. Just add the image to the PR in static/samples/

jasondunsmore:
Sure. I have some samples cut out, so I will try to take a photo today.

jasondunsmore:
I see that pytest is failing because the svg file references the URL of my local boxes.py instance instead of https://boxes.hackerspace-bamberg.de, but I'm not sure how to fix it.

florianfesti:
Yeah, the CI is more than a bit finicky. Fixed the PR. But the CI and its docs needs a bit more polishing to make this easier for  contributors.

florianfesti:
Anyway, thanks a lot for the generator!

florianfesti:
You you mind posting the output of `python -VV`. There is some difference between the output depending on the Python version and/or operating system. You files had a space before `/>` while the ones in the example dir have a space before `/>` while yours does not. And I am trying to track these differences down to get rid of them ad some point in the future...

florianfesti:
@Rotzbua Any idea about this?

jasondunsmore:
> Would you mind posting the output of `python -VV`? There is some difference between the output depending on the Python version and/or operating system. The files in the example dir have a space before `/>` while yours does not. And I am trying to track these differences down to get rid of them ad some point in the future...

```
$ python3 -VV
Python 3.11.2 (main, Sep 14 2024, 03:00:30) [GCC 12.2.0]
```

PR #738, opened 2024-12-26
New Generator Fatballfeeder

mrab:
- Added generator fatballfeeder
- Added translations (de, en)
- Added sample image


florianfesti:
This is a really nice generator and also build. I did find a few minor issues in the code that I already fixed.

I am currently working on a hackaday.io blog post looking back on the changes of 2024. This would make a nice addition there. So I am trying to get this merged this year.

Two things I still have in mind but are not really needed to get this merged:

I wonder if the poles should be mentioned in the "Other parts needed for assembly" and if this could make use of the [AngledCutJig](https://boxes.hackerspace-bamberg.de/AngledCutJig?language=en). May be by printing out the angles needed.

Anyway, as soon as we agree the generator is ready to be merged I'll squash all the commits and update the example SVG to make the test suite happy.

mrab:
Thanks for the kind feedback!

You're right, the poles should be mentioned. That slipped through as this section has been changed during the build (the wire mesh is already a bugfix release). I'll add that tomorrow morning (need to get some sleep now, it's congress time).

I wasn't aware of the angle jig. I reckon that it's not really needed unless someone wants to go into mass production or takes the angles very serious. I built two feeders just using sandpaper and a carving knife. The lack of precision was covered by paint. One would need jigs for four different angles, balcony wall, 2x roof tile and the support piece. The latter is also very shallow, so the laser would need to burn through a lot of material.

florianfesti:
OK, looks good so far. I have the sneaking suspicion that the code can be simplified at a few places but let's not worry about that now. Same for the idea to have the poles and mesh made out of laser cut pieces.

If you don't plan to do something right now we can just merge the PR as is.

mrab:
You're right, there is some code duplication, I didn't clean it up yet as the expressions are quite short. Extracting into functions seemed to be an overkill and I didn't find a proper / pythonic way to initialize the common values as member variables. In the ctor the arguments are not yet parsed and in render() the members should have been declared / initialized already in the ctor. 

There are currently no plans to cut the poles or the mesh out of sheet material. But I'll definitely consider that if I'll build another one. Another option to avoid the mesh would be to put in more poles.

florianfesti:
OK, let keep it as is. Thanks a lot for this very good looking addition. Great the Bamberg Backspace in the Soco assembly from me.

florianfesti:
Ahhh, forgot to move the generator out of "Unstable". Done now.

Issue #742, opened 2025-01-06
UniversalBox with hinge eye produce error


werner4711:
**Describe the bug**
Open https://boxes.hackerspace-bamberg.de/UniversalBox?language=en
Go to UniversalBox Settings -> top_edge and in drop down "Straight edge with hinge eye" 
Click generate.

**Expected behavior**
A generated SVG preview.
But got:
An error occurred! 'UniversalBox' object has no attribute 'grip_percentage'

**Additional context**
grip_percentage found in section "Settings for Hinges and HingePins".
default value 0 
When add any number than 0 same happens. 
Removing the value result in "An error occurred! argument --Hinge_grip_percentage: invalid int value: ''"

gpsqueeek:
This seems to be due to commit d4b238c (jan 3rd 2025) on the file edges.py, in the cases where the hinge setting "outset" is checked. I am not sure yet how to solve this without breaking the previous change at the same time...

PR #746, opened 2025-01-09
Kamishibai

gpsqueeek:
Hello,

Here is my first contribution... Quite a big one actually, at least I spent quite some time to improve my prototypes before submitting !
Please let me know if anything is wrong, so I can improve it.

Thanks for your huge work ! \o/

Have a nice day...

Squeeek


gpsqueeek:
It looks like issue #742 is causing a failure here (not sure it is the only one, I am still new to github...)

florianfesti:
This looks gorgeous!

I understand this is for paper which is not super strong. Still a few places look pretty fragile. E.g. the `top handle` pieces or the `front` and `back` which are almost cut to pieces by the screws.

Note that there is currently a way to add screws to the regular finger joints. This is currently not used anywhere but the code might still work. You need to pass `bedBolts` and `bedBoltSettings` either to the edge itself or to `rectangularWall`.

There is #36 but there hasn't been much movement lately...

florianfesti:
Added thumbnail and updated the example SVG. Is there anything left to do? Otherwise I'd just squash the commits and just merge.

gpsqueeek:
I was planning to add an extra option for extra holes for feet (for the screwed version) because the current version might fall without these ; probably also something for the doors in this case. I'll let you know as soon as I did that, I think I can do it within one week

gpsqueeek:
Oops looks like I still can't manage git correctly, I deleted the kamishibai.py file 😱
I'll try to restore it with my modifications 🫣

florianfesti:
`git reflog` and `git reset` are your friends.

gpsqueeek:
Hello, I built an A4 version with the new elements (extra door locks for the bigger sizes, optional extra feet for the doors), they seem to work fine. I think it is ready for integration now.
Any advice on how to make the tests pass ? My generated svg (from a clean and up-to-date repo) seems to be different, I don't know why... 🤔

florianfesti:
Don't worry. I can take care of that. This probably needs to be documented properly. The SVGs for the test are generated slightly different as they need to strip things like the generation date.

I usually just run `pre-commit run` and the copy the file from `tests/data/` to `examples/`. Probably need to make this a bit more accessible.

florianfesti:
The screw slots at the front panels look wrong and too close to the bottom lock.

gpsqueeek:
I did the pre-commit run trick and it worked fine too ! I hope my next contributions need less work for you, but you give the right advice for that, thanks a lot !

florianfesti:
Disabling the screws by default doesn't really solve the issue. Or may be I don't understand what is getting screwed on there. To me it look like the bottom of the front panels should not be screwed to anything.

gpsqueeek:
The screws are meant for the case where you prefer to screw the box instead of gluing or inserting the pieces together with a mallet. The counterpart is that you have to add feet under the box ; but this also means that the doors need feet if you don't want to hols the box while in use (opened in front) else the box may fall forward.
The other issue is that 3 mm (the default wood thickness) might be a bit thin as the box is often quite big (I use 5 mm) ; furthermore, the screw nut will probably too thick for the extra locking mechanism anyway. I will add more information in the description for that purpose, and also mounting instructions (should I add pictures for that ?).

gpsqueeek:
I cannot make the CI tests pass again, it seems to be linked to the multiline description... I tried quite a lot of fixes already but cannot find out the exact issue...

gpsqueeek:
Yippee ! I needed to stage my example file before I run the script...

florianfesti:
OK, decided to just squash everything. I hope this is in your interest.

PR #751, opened 2025-01-25
added none to --top parameter choices list

Korchy:
In the --top parameter of the roundedbox generator the value of "none" is set as the default value, but is not included in the choices list. So the argparse exists with the error 
app.py: error: argument --top: invalid choice: 'none' (choose from 'closed', 'hole', 'lid')
if the default "none" value is used when calling the generator execution.
Adding "none" to the choices removes the error.


florianfesti:
I fixed this slightly differently in e80ad1c42ac500a3ae6cecb0d767f7f0c3ab3109

Still a nice catch. Thanks for the report and investigation!

PR #752, opened 2025-01-26
fixed --top parameter default value in roundedregularbox generator

Korchy:
Same fix as in the commit e80ad1c (https://github.com/florianfesti/boxes/commit/e80ad1c42ac500a3ae6cecb0d767f7f0c3ab3109) but for the roundedregularbox generator.


florianfesti:
OK, I should have thought about that one myself. Thanks for the patch!

PR #759, opened 2025-02-23
Improvements to Gridfinity Base

maihde:
This PR provides multiple improvements, three are very straight forward:

- allow setting height to zero to disable creation of walls (i.e. you only want the grid base).
- set the openings to have a radius that is consistent with the Gridfinity specification and more asthetically pleasing in my opinion
- when `cut_pads` is turned on, also cut magnet holes by default.  I could see an argument to add a boolean `cut_pads_mag` which is off by default but I refrained to keep the command-line more concise.

The next is intended to make it easy to use boxes.py to build grids that drop into existing areas with a fixed size like drawers.  Instead of providing the number of grids in the `x` and `y` direction you can instead specify the full size (in mm) in the `sx` and `xy` similar to (https://laserbase.makershop.ch/).

I made a few design decisions here:
- the default behavior is consistent with the existing command-line for boxes.py, so anyone creating Gridfinity trays already won't see a change
- if `x, y, sx, sy` are all provided then the minimum size of the box is specified by `x and `y` (in other words, you are always guaranted `x` by `y` grids).  If `sx` and `sy` are larger than the mininum size, then the grid is centered in the larger base.
- if `x` and `y` are set to zero then the maximum number of grids is produced to fit within the provided `sx` and `sy`

The next change allows creating grids that are larger than a single work piece.  This is similar to how https://laserbase.makershop.ch/Split works (but better).  In the makershop version, splits are made arbitrarily.  In my implementation, the splits are made down the center between grids and joined with a dovetail joint.  A future commit will allow turning the dovetail off and using a straight edge.

In addition, when run in split mode it will also automatically split the box walls and base.  I'd really like the box base to use a deeper dovetail, but couldn't figure out how to change dovetail settings mid-render.  Help would be appreciated here. In addition, when running in split mode it won't generate a lid.


florianfesti:
This is quite a bit of changes. I will try to get them reviewed soonish, but it's currently a bit difficult to find the time.

I wonder if the magnets holes in the feet should have matching magnet holes in the bottom plate.

maihde:
Thanks for the feedback!  I will make those change

Regarding the comment about adding magnet holes in the base.  I'm happy to
do that, but I think it might be of limited utility.  I think the more
common use (which happens to be mine) is using the baseplate by itself
within the metal drawer of a tool chest.  If you are using a laser cut box,
it's probably easier and cheaper to put a thin sheet of metal in the bottom
of the box under the base, rather than trying to glue magnets into the base
hole.  If you'd like me to add it - just let me know!

~Michael

florianfesti:
OK; that's a valid point. I guess it shows that I have not actually used Gridfinity myself. Adding a few magnets to a box is a very different thing than filling a whole base plate with magnets. Don't bother with the magnet holes for now.

maihde:
@florianfesti I've pushed the changes requested.  There are various ways to use Gridfinity.  Here is one example, where I've used the modified version of boxes to create six panels for a desk drawer.
unnamed

florianfesti:
Looks very neat!

In my eyes we can just merge this now unless you want to clean up the commits and merge the fixes into the features.

maihde:
I'm fine with the commits as-is; so if you are okay with it, you can merge
at your convenience.  Thanks for the review and making a great piece of
software.

florianfesti:
Thanks for all the additions!

PR #771, opened 2025-03-24
Various Improvements to Gridfinity Tray

maihde:
Commit messages have details.  Changes summarized are:

- Fix confusing help message on webpage that height is "inner height"
- Support defining height in terms of Gridfinity Units (i.e. 7mm per U)
- Support easily making stackable boxes compatible with Gridfinity
- Enable turning off pad generation
- Cut pads with gridfinity compatible radius
- Cut pads with holes for magnets

See https://www.thingiverse.com/thing:6979985 for example

### DIFF
--- a/boxes/generators/gridfinitytraylayout.py
+++ b/boxes/generators/gridfinitytraylayout.py
@@ -1,7 +1,7 @@
 import argparse
 
 import boxes
-from boxes import Boxes, lids, restore
+from boxes import Boxes, lids, restore, boolarg
 from boxes.Color import Color
 from boxes.generators.traylayout import TrayLayout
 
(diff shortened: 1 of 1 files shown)


boxes/generators/gridfinitytraylayout.py
```diff
@@ -111,6 +112,34 @@ def render(self):
...
+            #
+            # The goal of the above code is that a stack of laser cut boxes and a stack
+            # of 3d printed boxes will end up the same total height
+
+            self.h = (int(self.h[0:-1]) * 7) - self.thickness
+
+            if self.stacking:
+                self.h += 3.69
+        else:
+            raise ValueError("--h must be a number of a number followed by 'u'")
```
florianfesti, line 174: Guess this should be _or_ a number followed by ...

florianfesti:
Thanks a lot for all this work! Took me a bit to get through it.

PR #774, opened 2025-03-30
Misc documentation improvements

rudis:
Hi,

a few patches to (hopefully) make the API documentation a bit clearer.

I don't have much experience with boxes.py so please check if my assumptions are correct, for example "documentation: part move direction is unrelated to draw direction".

The static patch might also fix the broken images on https://florianfesti.github.io/boxes/html/README.html

Best,
Simon


florianfesti:
Thanks a lot. There is just no replacement to someone else looking at the docs.

PR #780, opened 2025-04-04
New Generator: Sliding Lid Box

felis-blue:
I added a new generator for a box with a lid that slides between two rails.

* The sliding lid rests on the lower rails and is kept in place by the upper rails.
* When using inner measurements, the height goes from bottom to lid, the lower rail is inside of the inner height.
* The width of the rail can be adjusted, wider rails are more stable but make the opening smaller.
* The horizontal margin makes the lid slightly narrower to prevent the lid from jamming.
* The vertical margin makes the gap between the rails larger to let the lid slide more easily.

Example Images:
IMG_1985
IMG_1986


florianfesti:
This box is not completely unlike both https://boxes.hackerspace-bamberg.de/CompartmentBox and https://boxes.hackerspace-bamberg.de/CardBox . I wonder if they shouldn't be merged all into one generator. May be on some cold winter day...

This looks pretty good already. I wonder if it should have a front lip similar to the other two. This allows to grip the lid and pull it open. It can also just go into the opening between the side rails. The CompoundEdge allows to put "eFe" along the edge to match the rails.

florianfesti:
After sleeping over it: May be just gluing on a rectangle actually looks better.

felis-blue:
I thought about adding a front lip to the lid, but the way the box is constructed, this would either be only one "thickness" wide or stick up over the rest of the box. Especially with thinner material, I feel like the lip would be a bit unstable.

I have "manually" stitched together this box a few times before writing this generator, and I usually add a finger hole to the lid, to make it easier to slide in and out. I could maybe try to add an optional finger hole setting like in https://boxes.hackerspace-bamberg.de/CompartmentBox?

florianfesti:
Finger hole also works. I wonder if just gluing a one layer grip on top is even better. If you really don't like gluing you could "nail" it in place with some small two thickness long bars - with or without a head. Guess I should add such connectors to the parts at some point...

felis-blue:
I have now added a hole and a lip option, for better lid handling options. I have also switched the back rail from f to F, to make it more stable.

florianfesti:
Well, just a piece of wood glued flat onto the lid. Anyway I can add this later on myself.

If you don't mind I'll crop the images and add them to the generator. Then this is ready for merging.

florianfesti:
Here they are.

florianfesti:
And merged! Thanks for your work!

PR #781, opened 2025-04-08
Misc documentation improvements

rudis:
Hi,

a few more minor updates to the documentation.

Best,
Simon

PS: I couldn't figure out how `rectangularWall()`'s `ignore_widths` is supposed to work. In the end I didn't need it but a short explanation in the doc string would be nice.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -2432,6 +2432,11 @@ class Boxes:
         """
         Rectangular wall for all kind of box like objects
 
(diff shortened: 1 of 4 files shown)


florianfesti:
Thanks a lot for the improvements. I'll try to come up with an explanation for the `ignore_widths`

florianfesti:
Does this make things clear?

rudis:
Yes, that makes it clear. Thank you!

PR #782, opened 2025-04-16
desksign: Support multiple text rows

opticron:
This adds support for multiple text rows by evaluating backslash escapes for newlines and adding a parameter to control the vertical offset of text.


opticron:
Clearly I need to update a test or two. I'll get to that tonight.

florianfesti:
Just add `\n` in the default value of label and you'll get a textarea as input field and multiple lines will work automatically (hopefully).

This `label = bytes(self.label, 'utf-8').decode('unicode_escape')` business makes me nervous.

The test suite needs a bit of docs still. You need to have your changes committed or at least added in git. You can then copy the new file over from `tests/data/` to `examples/`.

florianfesti:
Ok, fixed this up myself. Still thanks for the patch!

Good that I did as I also found out that the preview did not update for textareas.

PR #783, opened 2025-04-18
Add option to remove front lip on hobby case

dstotz:
New option for inset_shelves added to hobby case to allow for optionally removing the small lip at the front that insets the shelves. New option is set to true by default to preserve existing behavior


florianfesti:
Thanks for the patch! Did also clean up the generator a bit as I was already looking at it...

PR #786, opened 2025-04-22
New Generator: Apothecary Drawer Box

dstotz:
- Adds a new generator for Apothecary style drawer box and optional drawers using existing generators
- Create Docker Compose file for easier local development
- Update documentation
- New labelDimensions option for rectangularWall to automatically label dimensions
- Updated ABox and DividerTray generators to use new labelDimensions option

## Note
This generator has not been used yet as I am currently waiting on the material to arrive. I will update the image and make any tweaks necessary for issues that may appear during cutting. Opening the PR now to get early code review / feedback


florianfesti:
This really should be 3 different PRs. Anyway let's have a look at the commits one by one.

While I think adding the dimensions to the label is useful for some people I am not a fan of adding a new parameter to `rectangularWall`. This adds just to much code for a small feature and leaves it incomplete almost everywhere. I'd rather see the `labels` standard parameter growing an option to enable this for all pieces. Probably by making it a choice with `none`, `description`, `measurements` and `both`. Just keep the default as `description` and you don't even need to touch any example SVGs.

I am all for making the use of the Dockerfile for development purposes easier. This would just be merged if it was a separate PR.

Now to the main course. We totally want something like this generator. Having an actual image would be appreciated, though.
I wonder if there is a more elegant way of solving the finger joints. If you force a even number of fingers you can just do all uneven fingers. As the edges of adjacent pieces are in opposite directions they should fit together. That way you only need one type of edge. The `Platonic` generator does something not completely unlike this. This feature should probably added to the `FingerJointEdge` as a general feature enabled by a private parameter in the `FingerJointSettings` similar to the `angle`.

While not a requirement to be merged an option to just add rails similar to the `HobbyCase` generator might be a nice addition.

dstotz:
@florianfesti thanks for the feedback! I will go ahead and break this out into multiple changes so we can work on each independently and rebase as needed. Would you prefer to have this PR get closed and I'll open a new one once all dependent changes are merged? Or are you okay with me leaving this PR open and I'll reference the other PR's in a comment so they are linked?

For the labels, I totally understand. I was hoping quick and dirty would do the trick, but in the end it's not a good look. I'll give your recommendation a shot.

As for the alternating finger joints, I will definitely take a look into that. If I am unable to make that work without too much difficulty, would you be open to keeping the joint implementation in-line as-is and then creating an issue for a future todo to refactor it and make it reusable?

And I do have an image of a smaller test print I did so you can see how it looks. I was just going to wait until I printed the full size 3x4 setup before submitting the real image, but if you think this is good enough, I can go ahead and crop and add this one now.
IMG_1780

dstotz:
Relies on #794 for passing label_formats arg to sub-generators

dstotz:
@florianfesti I tried out the platonic implementation but it didn't work for my use case since I don't want them to actually interlock, I want them to intentionally alternate skipping a joint.

I ended up going with the `FingerJointSettings` approach you recommended which really let me dry up the implementation and make this reusable.

I'm going to move this PR off of draft and it is ready for review!

PR #787, opened 2025-04-24
Update zh_CN.po

hznupeter:


florianfesti:
This looks all good (not that I speak a word of Chinese). Just drop a comment if you are as far as you want to get and I'll handle the other details.

hznupeter:
translate to chinese，fix some translate errors

florianfesti:
Thanks for your work! The translation should show up on the web instance in a few minutes.

PR #788, opened 2025-04-28
Minor Fixes to GridfinityBase and GridfinityTrayLayout

maihde:
- The GridfinityBase can incorrectly split parts in some cases because the split was not accounting for the pad. This was introduced when adding split functionality.
- GridfinityTrayLayout etching of pad locations will write the etching multiple overlapping times when the size of the tray is 1 or 2 grid units in either the x or y direction.


florianfesti:
Would you mind switching to a consistent use of white space?

maihde:
Happy to!  Sorry about that.

maihde:
Fixed.

florianfesti:
Thanks for the fixes. Nifty to use sets there.

PR #790, opened 2025-04-29
feat: support gridfinitytray compatible with both laser and 3d printed bases

maihde:
The Gridfinity specification establishes that one height unit is 7mm, so a 6U box would be exactly 42mm high.  When a laser cut box sits on a laser cut base, the box wall height is the total height minus the thickness of the material being cut.  When a laser cut box sites on a 3D printed base, the base itself is 4.65mm (per the Gridfinity) specification so the box walls need to be shorter.

This can be seen in the following pictures:

A 6u (42mm) tall tray with stacking header (3.7mm) is a total of 45.7mm either 3d printed or laser.

image

A 3d printed tray can sit on either a 3d printed base or a laser base and will still be the correct height.  However, a laser cut tray can only sit on a laser cut base and be the correct height.  If a laser cut tray is on a 3d printed base it will sit too high.

image

The expectation is that most users that laser cut trays are also laser cutting bases; thus, they would never experience this incompatibility.  To accomdate the case where a 3d printed base is used, this commit allows them to override the base height with a custom base height.  

image

This same behavior could also be accomplished without any change by specifying height in mm instead of Gridfinity units; however, for Gridfinity, it's more natural to terms of unit height is more natural.

### DIFF
--- a/boxes/generators/gridfinitytraylayout.py
+++ b/boxes/generators/gridfinitytraylayout.py
@@ -44,6 +44,7 @@ this compartment.
         self.argparser.add_argument("--pad_radius", type=float, default=0.8, help="The corner radius for each grid opening.  Typical is 0.8,")
         self.argparser.add_argument("--cut_pads_mag_diameter", type=float, default=6.5, help="if pads are cut add holes for magnets. Typical is 6.5, zero to disable,")
         self.argparser.add_argument("--cut_pads_mag_offset", type=float, default=7.75, help="if magnet hole offset from pitch corners.  Typical is 7.75.")
+        self.argparser.add_argument("--base_thickness", type=float, default=0.0, help="the thickness of base the box will sit upon.  0 to use the material thickness, 4.65 for a standard Gridfinity 3D printed base")
         if self.UI == "web":
             self.argparser.add_argument("--layout", type=str, help="You can hand edit this before generating", default="\n");
         else:
@@ -166,7 +167,11 @@ this compartment.
             # The goal of the above code is that a stack of laser cut boxes and a stack
             # of 3d printed boxes will end up the same total height
 
-            self.h = (int(self.h[0:-1]) * 7) - self.thickness
+            self.h = (int(self.h[0:-1]) * 7)
+            if self.base_thickness == 0.0:
+                self.h -= self.thickness
+            else:
+                self.h -= self.base_thickness
 
             if self.stacking:
                 self.h += 3.69


florianfesti:
Hmm, Gridfinity boxes can also stack on top of each other (for some boxes), right?

I wonder if it would be better to make the default so that the above the base height is the same as the laser printed one. I think the main use case is mixing 3D printed and laser cut boxes on whatever base. With the base being 4.65mm high the boxes are most likely sitting on the base as using more than 4mm wall thickness seems a bit much.

maihde:
Yes - if you print/cut your boxes with the stacking-lip, then you can stack boxes.  Here is an example where I have both 3d-printed and laser-cut boxes stacked on a laser-cut base.  These boxes were cut with boxes.py with the defaults and a h=6u and stacking=1

image

In this configuration, both laser cut boxes and 3d print boxes stack to the same height at each layer because we are using a laser cut base.

Regarding what the correct default is, I don't think there is one that allows compatiblity with laser and 3d printed boxes on whatever base under the assumption that cuts are made from 3mm material.  Part of this is the difference in base/pad thickness between laser and 3d-printed, but part of it is the 3d-printed specification uses a tapered configuration that is impossible to replicate with a laser cutter.  Below is the specification for Gridfinity base/pads.

image

Because of this taper, 3d-printed boxes can never stack correctly on top of laser-cut boxes.  Obviously you can stack them, but they won't line up with a stack of only 3d-printed or only laser-cut boxes so columns need to be organized correctly.

My assumptions, based on my use-case:

a) if you are laser cutting boxes, you are also likely laser cutting bases.  This is because it is substantially faster to cut bases rather than 3d-printing them.
b) when making boxes they are for a specific base. So my primary concern is compatibility with the chosen base, not moving boxes between bases.

With the current defaults (`base_height=thickness`), someone that cuts boxes and bases with boxes.py will end up with results that are the correct height and compatible with 3d-printed boxes (as in the picture above).

I agree with you.  If the pads/feet are cut from 4mm or 5mm material, then the laser cut boxes would be effectively compatible in all configurations, but I don't think most people would want to cut the box walls with 4 or 5mm thick material.

Anyways, I'm happy to make the default whatever you think best.  For me, the current behavior is most intuitive, but I'm a sample size of one :).

Finally - thanks for boxes.py and incorporating my previous changes and considering this and future contributions.  I really like boxes.py and Gridfinity and I'd like to see boxes.py be the best Gridfinity-laser solution!

florianfesti:
Well, I have not really used Gridfinity myself yet. Let's just go with your patch for now. It will at least allow people to make it work for them.

PR #798, opened 2025-05-27
Lazy susan

mtintes:
Hi I am trying to make a little box setup for a Lazy Susan. One of [these](https://www.homedepot.com/p/Rev-A-Shelf-2-Shelf-White-28-in-Lazy-Susan-Pie-Cut-Polymer-6942-28-11-52/312843737) things in case there is confusion.

I got pretty far, except I can't quite figure out how to make the curved walls. I've tried a few examples and think I need a polygonwall? but I am unclear how the parameters to that works even after experimenting.

Maybe someone could help me out? The second and third chunks of the svg are, I think, the right length but the inside of them needs to be flexible going the horizontal direction so they can wrap the inside and outside of the curve on the first shape. I spent a week getting this far and I'm hoping there is an easy way that I'm just not getting.

Any help would be appreciated. TY in advance. 

some things I have tried:
followed heart box example (this is kinda the same shape)
followed alledges example, but was getting an error about thickness when using self.edges["x"](h)
polygonWall(borders from heart box) - I still couldn't figure out what the parameters do there.
Searching for a method with "flex" or something in the name
read a ton of the api docs and didn't see anything
flex2d was a both way pattern instead of one direction.


florianfesti:
The link to home depot seems to be blocked outside of the US.

For something mechanical I'd probably try to avoid flex. Currently flex does not have any finger joints. So it is not really a connection. Or at least combine with straight edges in between. What about just using a RegularBox or RoundedRegularBox?

The `borders` parameter is a list if alternating length and angle starting with length.
Length can be a just a number or a tuple of a length and a number of [tabs](https://florianfesti.github.io/boxes/html/usermanual.html#tabs). The angle can either be just a number or a tuple of angle and radius. The later is what you need for flex. Length can be zero if you want to angles or radii to meet.

mtintes:
sorry about the link. I should have just used Wikipedia [here](https://en.wikipedia.org/wiki/Lazy_Susan). 

Just for clarity I added a badly drawn example of what I was going for. I don't know that the box examples can bend around the shape quite as nicely. I think you are right about the strength of the flex walls. My plan was to run a few tests and if they were too weak then I would also add a rail around the top and those side walls, the orange and pink walls, would bear the weight of that. Then the flex walls would just be decorative to make it look nice and a little wood glue would hold the top and bottom to the rest of the frame. 
image

florianfesti:
So these are Boxes you want onto of the Lazy Susan? Or is this the base? Why is this only a quarter circle? Do you have a rough size estimate?

florianfesti:
OK, here a quick sketch using polygonWall(s). You might want to use more straight walls. May be every 30° or so.

mtintes:
> So these are Boxes you want onto of the Lazy Susan? Or is this the base? Why is this only a quarter circle? Do you have a rough size estimate?

yup you got it. It would sit on the lazy Susan and would have a floor to the box. You could put a few of these boxes on the lazy Susan for organizing. I made it so you can give the box an angle and that is how much of the circle (or arc) it will make. I just defaulted to a quarter circle. We put tall bottles of oils and stuff in ours and the centripital force knocks them over and this could solve that.


> OK, here a quick sketch using polygonWall(s). You might want to use more straight walls. May be every 30° or so.

Oh I like the idea of the flat walls, that could maybe simplify it. I like the look of the flex walls, but I just need to make something that works and that would still look pretty nice I think. It would probably give more strength so that would be good.

I'll give it a shot.

florianfesti:
Is this finished already or do you need some help with anything?

PtitLutinn:
Hi, I was going to make the same thing as you to put inside a empty filament spool. I did a first test with openscad for the parametric shape and inkscape for the line for the flex wall. I cut it this afternoon and it's holding pretty well. I will have a look at your code if you haven't figured out how to do the flex wall because I really want it. I thank about the hexagonal wall but my box is to small for it render ok. and I like the look of the flex wall. 

IMG_20250731_195920
IMG_20250731_204805

PtitLutinn:
i have tested the code, they are some issues (see picture), I will try to resolve them by doing a fork of mtintes branch (don't know if it's the correct way to do it thought ^^')
I will also try to add the exterior parameter which is tricky because of the circle involved (I have made the calculs before doing my test since it's something I really need) 

https://cdn.discordapp.com/attachments/1328733687327490048/1401909423470940362/rn_image_picker_lib_temp_8b56587c-bff3-4af6-9c33-ddf2824494f2.jpg?ex=689d3224&is=689be0a4&hm=dff193d1f83be08df1f98dacc2fc8f3942cd65ce8e1722bcfe1667574726d77d&

PtitLutinn:
this one doesn't work. the size of the wall is ok but it shift the fingers joints on the midle of it and the added height endend up on the top and the bottom of the wall when we want it only on the bottom. I think the correct thing is to use the 'E' edge (see here : https://florianfesti.github.io/boxes/html/api_edges.html) but I couldn't make it work, If you find the correct way to use it

PR #804, opened 2025-06-04
Add engraving frame generator

gpsqueeek:
New generator for a wooden engraving and its frame. The frame can be either hanging (horizontally or vertically) or standing on its optional feet. An optional extra frame can be glued on top to hide the burnt fingers.


florianfesti:
Sorry, for the late reply.

This looks really nice. Merged.

PR #806, opened 2025-06-10
Improve Chinese translation

hznupeter:
Improve Chinese translation. Almost done!


florianfesti:
Thanks for all the work!

Issue #755, opened 2025-02-01
Sphereoid lamp


JolienF:
Hey,

So I made this lampshade (its not perfect yet, but the prototypes are looking good). Its a sphere like shape that can be cut from flexible material like cardboard or plastic sheets. The amount of gores (https://en.wikipedia.org/wiki/Gore_(segment)) is variable. The pattern is not generated (yet). 

So my question is would this be a project to put on boxes.py? It's not the  standard wooden boxy type thing. But it is generated and very lasercuttable
I'd love to give it a go to integrate it in this codebase, but I could definitely use some tips. 

![Image](https://github.com/user-attachments/assets/88606ea6-a3d0-4659-a144-5f983405c447)

![Image](https://github.com/user-attachments/assets/e1352a18-6b60-43ee-814e-9fe076646f8d)

JolienF:
![Image](https://github.com/user-attachments/assets/8706e647-3731-41f7-bc6f-8feb63e51143) this is the type of pattern it generates

florianfesti:
Yes, of course. There are a lot of things in there already that are not quite boxes.

ccrome:
Love it!  I'm happy to give it a try!

florianfesti:
A little more gore, what's not to like...?

JolienF:
Alright, I'll get started then!

JolienF:
> Love it! I'm happy to give it a try!

I'd like to try myself first, if you don't mind. I do appreciate some help though

florianfesti:
Generally Boxes.py is not that suited for rounded shapes. I does offer very little geometry support. The gears module uses `drawPoints()` which does kerf correction and draws a polyline from coordinates instead of lengths and turns.

An alternative could be using `curveTo` which draws Bezier curves but looks like it is missing burn correction although it is in a place in the API where it should actually do that. If this gets an issue I can fix that, though.

JolienF:
@florianfesti what do you reccomend for circles? I use some halfcircles, but the disc and circles in boxes don't have a kerf or an option to create only half a circle.

florianfesti:
You can just use `.corner` and `.polyline`. See https://florianfesti.github.io/boxes/html/api_burn.html

https://florianfesti.github.io/boxes/html/boxes.html#boxes.Boxes.polyline

https://florianfesti.github.io/boxes/html/boxes.html#boxes.Boxes.corner

PR #813, opened 2025-06-28
Sphere generator

JolienF:
So finally a proper version of the sphere generator. It works nicely, but I'm sure it could be improved, so feedback is welcome. 
The generator draws the gores that make up a sphere. It can be assembled with tabs and even has optional scoring lines and hooks to hang the sphere. It does not include a generator for hole patterns in the lamp (see the picture of https://github.com/florianfesti/boxes/issues/755)
I plan on making a couple of lamps this week and then taking some nice pictures (currently I only have ones with lots of burnmarks)


JolienF:
So what can I expect at this point? haha, I did the big scary pull request and nothing happened for 4 days. I know its a busy time just before summer, so I don't mean to push, but I'd like to know what to expect. Is there anyone who could just give me an estimate on how long these things usually take?

ccrome:
> So what can I expect at this point? haha, I did the big scary pull request and nothing happened for 4 days. I know its a busy time just before summer, so I don't mean to push, but I'd like to know what to expect. Is there anyone who could just give me an estimate on how long these things usually take?

Florian is normally *extremely* responsive.  Just give him a few days.

florianfesti:
"extremely responsive" is not the words I would use especially when it comes to PR reviews. I am behind with those - by a lot.

OK, I am not checking the math on this one. But playing around with some values shows a few things of note.

* Choosing a `top_hole_radius` or `bottom_hole_radius` bigger than `sphere_radius` breaks the generator. You can either just clip the values or raise a `ValueError` that then will be shown in the UI.
* I'd expect the top and bottom edge to be the same for the same radius. But the bottom edge is always straight
* If the length of the tabs goes below `tab_width` the tabs overlap and get cut off. Not a big deal but may be reduce the number of tabs for very ring like "spheres".

Overall this is a very nice generator. The minor issues above are not a requirement for this to go in, although the first one is really easy and probably worth doing. Just put a comment when you have fixed the things you want to fix.

JolienF:
Well, that was way more difficult than I thought. The bottom edge is straight because I like the design of seeing the polygon on the bottom, I tried implementing a circular hole, and I got close but I couldn't get it to work. Maybe I'll try again in a month or so, but for now it would be nice to put it out there. The other points are implemented (the ValueErrors for the top_hole_radius, bottom_hole_radius and the tabs).

JolienF:
Also I'll be giving a talk at WHY2025 about how the maths work. I'd love to meet up if you happen to be there

florianfesti:
Added some of your pictures and squashed the commits. Thanks for your patience and you work!

Unfortunately I am not at WHY. May be your talk gets recorded?

JolienF:
Yes, it will probably be recorded. Thank you for doing boxes.py! I've been using it for years and it feels good to finally give something back.

PR #825, opened 2025-08-28
Add functions to import and draw paths from an SVG.

rwb27:
This adds a few utility functions that can load a path from an SVG file and draw them on a box. I have tested only two files, which worked - I'll attach the OFM logo here.

I have also attempted to make a more neutral SVG in Inkscape and in OpenSCAD - but unfortunately neither worked properly. I think that means these functions are unlikely to be very useful without further testing. Nonetheless I'll leave them here in case it's of use to anyone.

### DIFF
--- a/boxes/svg_import.py
+++ b/boxes/svg_import.py
@@ -1,0 +1,61 @@
+"""Allow paths to be imported from an SVG file into boxes.py.
+
+This is intended to make it easier to add custom cut-outs and engraving.
+Currently, only a single path can be imported. This is generally sufficient
+to import shapes drawn in OpenSCAD.
+"""
+
+import svg.path
+from lxml import etree
+from boxes.drawing import Context
+import numpy as np
+
+def draw_path_on_ctx(ctx: Context, path: svg.path.Path):
+    """Draw an SVG path into the current context.
+    
+    This transforms the points according to the current context, but otherwise
+    should just add the commands from the path into the output.
+    """
+    for seg in path:
+        if isinstance(seg, svg.path.Move):
+            ctx.move_to(seg.end.real, seg.end.imag)
+        if isinstance(seg, svg.path.Close) or isinstance(seg, svg.path.Line):
+            ctx.line_to(seg.end.real, seg.end.imag)
+        elif isinstance(seg, svg.path.CubicBezier):
+            ctx.curve_to(
+                seg.control1.real,
+                seg.control1.imag,
+                seg.control2.real,
+                seg.control2.imag,
+                seg.end.real,
+                seg.end.imag,
+            )
+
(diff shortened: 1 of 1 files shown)


rwb27:
Here are some example SVG files I've used that work.
[logo_emblem](https://github.com/user-attachments/assets/df1c0605-80a5-4d58-99ed-fa7496bc1e74)
[shelf_cutout](https://github.com/user-attachments/assets/7e72dad7-666c-4afd-8661-3742fb9e2ce7)
[footprint](https://github.com/user-attachments/assets/0df3114f-4c4f-45bb-ab1b-9cb3d24e2bda)
[footprint_cable_tidies](https://github.com/user-attachments/assets/7a7439c1-2486-4b28-adb4-2aeb6a0b933a)

rwb27:
Here are some examples that failed.
[openscad_smiley](https://github.com/user-attachments/assets/de5c1c35-6428-4719-aace-629cf214d38c)
[smiley](https://github.com/user-attachments/assets/2a59e349-3af9-40bf-9278-1c35229dcc79)

MartinScharrer:
> Here are some examples that failed. [openscad_smiley](https://github.com/user-attachments/assets/de5c1c35-6428-4719-aace-629cf214d38c) [smiley](https://github.com/user-attachments/assets/2a59e349-3af9-40bf-9278-1c35229dcc79)

Hi, they fail how? The resulting image is incorrect?

I did some similar code in my fork this week (to draw a logo and special cutouts imported from SVG) and I noticed that having multiple `z` commands in the path does not work, i.e. generates an incorrect image. You need to close the path so far every time a `z` or `Z` command is encountered. This should fix the issue.

I added this indirectly by calling ctx.stroke on `Z` but calling a close-path command might be better. Feel free to compare my code at https://github.com/MartinScharrer/boxes/blob/51598cef72447fa54ddd7c31ee17fd74fa8bed4d/boxes/generators/queentransportbox.py#L121

MartinScharrer:
The issue #847 might be relevant here. Maybe this is the reason some SVGs fail to render properly. Curves can then changed to edge polylines! 

I fixed it as PR #848.

florianfesti:
OK; I finally looked into this. Looks like the issue with the `openscad_smiley` are already resolved with the fixes from @MartinScharrer . The other `smiley` not so much, though. Reason is that the code currently does not support arcs (`svg.path.Arc`) which are used by that file. I am kindof OK, to not support the full SVG feature set. But arc are a base feature of paths. So we need to add support. The issue is that the Boxes.py backend currently only supports simple arc while the SVG arc are elliptical (or can be). So we'd need support for that in the backends first.

We can ofc argue what other SVG feature we want to be supported. But in my mind this is for generators shipping their own graphics and not for users adding graphics. That's something they can do themselves with Inksacpe. So we only need to make it easy enough for generator authors to get their SVG files in a supported form.

PR #827, opened 2025-09-03
Added an option for a latched lid to RoundedBox

rwb27:
This adds an additional value for `--top` which is `lid_with_latches`. Using this will add an extra plate that sits just inside (i.e. underneath) the lid, with four latches on it. The latches may be moved in and out using four M3 screws.

To assemble, the lid must be glued together with two smaller plates, to make a single unit. One of these plates is the inside of the top of the box (i.e. the part that's normally discarded when `top` is set to `lid` or `open`). That plate has four slots in it, where the screws pass through. This plate is glued directly on to the lid.

The second plate has four "fingers" cut out at the corners. This sits just inside the box, and should be glued onto the plate with the slots. Those two plates are the same size - it's a little tricky to get them lined up precisely, but this is important.

Once the three plates are glued together, the four M3x6mm screws pass through the lid (where there are large clearance slots for the screw heads), through the middle plate (where there are narrower slots that the shaft of the screw can pass through but not the head), and screw into the four fingers. If you slide the screws outwards, the fingers will lock the lid in place.

The holes for the screws need to be threaded before they will work - for my acrylic box I used a tap to cut the thread. Depending on the material it may be possible to self-tap, though that might work better with a different screw. I used M3x6mm screws and they worked neatly. I'm open to feedback on the best kind of screw to specify, though it would of course be easier to test ones I can find in my lab...

I've printed this in 3mm acrylic and it worked fairly well. I can't see why it wouldn't work in plywood/MDF. All the example photos are with wood materials, so I'm guessing if there were to be photos of this, they'd want to be similar?

### DIFF
--- a/boxes/Color.py
+++ b/boxes/Color.py
@@ -14,3 +14,6 @@ class Color:
     ANNOTATIONS = RED
     ETCHING = GREEN
     ETCHING_DEEP = CYAN
+    CUT = [  # Even steps from INNER_CUT to OUTER_CUT
+        [0.0, 0.0, 1.0 - i/4.0] for i in range(5)
+    ]
--- a/boxes/generators/roundedbox.py
+++ b/boxes/generators/roundedbox.py
@@ -124,6 +124,32 @@ With lid:
                     self.rectangularHole(0, 8*t, 1.1*t, 0.7*t),
                     self.rectangularHole(0, 10*t, 1.1*t, 0.7*t)))
 
+    @boxes.holeCol
+    def latch_holes(self):
+        """The holes cut out of one latch slider."""
+        t = self.thickness
+        self.rectangularHole(0, 1.5*t, t, 3*t, center_y=False)
+        self.hole(0, 9.5*t, d=5*t),
+
+    @boxes.holeCol
+    def latch_slider(self):
+        """The sliding piece that actuates the latch."""
+        t = self.thickness
+        self.rectangularHole(0, 0.55*t,
+                             6.95*t, 12.95*t,
+                             r=7*t, center_y=False,
+                             color=boxes.Color.CUT[1]),
+
+    @boxes.holeCol
+    def latch_cutout(self):
+        """The hole in which the latch_slider sits."""
+        t = self.thickness
+        self.rectangularHole(0, 0.5*t,
+                             7*t, 15*t,
+                             r=7*t, center_y=False,
+                             color=boxes.Color.CUT[2])
+
+
     def latches(self):
         t = self.thickness
         x, y, r = self.x, self.y, self.radius
@@ -142,12 +168,9 @@ With lid:
         self.latch_positions(
             lx, ly, r,
             lambda: (
-                self.rectangularHole(0, 1.5*t, t, 3*t, center_y=False),
-                self.rectangularHole(0, 0.55*t,
-                                     6.95*t, 12.95*t, r=7*t, center_y=False),
-                self.hole(0, 9.5*t, d=5*t),
-                self.rectangularHole(0, 0.5*t,
-                                     7*t, 15*t, r=7*t, center_y=False)))
+                self.latch_holes(),
+                self.latch_slider(),
+                self.latch_cutout()))
 
     def cb(self, nr):
         h = 0.5 * self.thickness
@@ -157,7 +180,17 @@ With lid:
             h += dh
             self.fingerHolesAt(0, h, l, 0)
 
+    def cut_order(self):
+        self.move(self.reference, 10, "down", before=True)
+        self.text("cut order:", self.reference + 25, 5,
+                        fontsize=6, align="middle center", color=boxes.Color.ANNOTATIONS)
+        for i, col in enumerate(boxes.Color.CUT[:3] + [boxes.Color.OUTER_CUT]):
+            self.text(f"{i}", self.reference + 50 + 10 * i, 5,
+                            fontsize=6, align="middle center", color=col)
+        self.move(self.reference, 10, "up", before=False)
+
     def render(self):
+        self.cut_order()
 
         _ = self.translations.gettext
         x, y, sh, r = self.x, self.y, self.sh, self.radius


florianfesti:
I like this a lot but I was pondering whether the latches can be done without vitamins. I tried to keep the latches flat so they can also be used on the bottom / with the box upside down. As it does not use round screws in the slots it works without an additional layer for guiding the latches.

I am on vacation this week, so this is still untested.

Edit: It also take the really great looking cross arrangement from the original and makes it part of the outside look.

florianfesti:
This needs to be refactored at some point to be usable for other generators with similar lids. But this is out of scope here. Let's get the RoundedBox right first and then think about where else this may be useful.

rwb27:
Thanks - that looks very neat! I did think about trying to do it screw-less, but concluded I couldn't come up with a design that was strong enough in the time I had available. The modified latches are, if I read it right, relying on a glued connection between the sliding plates on the bottom and the latches, so if that fails the latch can pop out. Probably that's not a problem (because we're already relying on a glued connection between the lip of the box and the wall), but I guess the latch-to-plate contact area is smaller. The sliding plates on the outside are neat - I didn't mention it, but I already use mine upside down (the screw heads are thinner than the 3mm acrylic): as you say, it's nice to be able to lift the box off, and have something sitting on the lid.

I wonder if it would be possible to modify it further, such that the latches hook onto the plate on both sides when they are in the "out" position? That would then mean we're not relying on a glued joint with the sliding plates. I might see if I can prototype that. I might also see if I can have a go at cutting this version - I'm traveling later this week but will see if I can try it out tomorrow.

rwb27:
I've implemented the suggestion above in https://github.com/florianfesti/boxes/pull/827/commits/e0ef44ca3906c9e0273a2490f79d8406e3194bf3. I think it's a fairly minor tweak: I've added hooks to the latches, and increased the length of the slots in the sliding plates to accommodate it. That's moved the edges of those sliding plates slightly closer to the edge of the lid, but I don't think they are problematically close. I've added a screenshot in case that is helpful.

Revised latch shape

Extended slider

I'll try to have a go at cutting this tomorrow, if I can.

florianfesti:
Funny enough that was what my first version of this piece looked like. And then I thought this is going to be too bulky. Turns out it is not.

rwb27:
I've printed this and it works nicely :) The clicking in and out works really nicely provided the lid is glued in place accurately - and the sliders are really good for helping locate the lid. Without the lid, the latches can be misaligned laterally, so they don't always click into place.

My first snag was that the sliders cut in the wrong order. I needed to use more layers in the laser cutter, so that it first cut out the holes in the sliders, then cut out the sliders, then cut out the lid piece. The problematic part is the one in the second screenshot above. I don't know if you have a standard way to do that - clearly we've used `@holeCol` already so the slider cuts before the lid, but I think we need a minimum of 3 and ideally 4 layers to get this to work reliably. We probably need to cut:
* The holes in the slider
* The outline of the slider
* The hole in the lid that goes around the slider
* The outside of the lid
Currently, the first three are all `@holeCol`. Additionally, sometimes I found that the slider ended up sitting under the lid and got etched by the laser when it cut out the hole in the lid. Perhaps that can be made less of a problem if we don't duplicate the path around it, but only cut the additional moon-shape part needed to make the slot the right length. I realise that's probably more complicated to code than the current version.

It might actually be simpler to move the sliders outside of the lid. That would be slightly wasteful of material, but would then only require two colours. Happy to be steered by you as to what is the "done thing".

Also, I glued on the lid after I'd assembled the sliders. That made it much easier to assemble the lid correctly (i.e. get the two plates lined up) but on my first attempt I also glued one of the sliders = game over. I snapped it trying to free the slider. Not quite sure what the solution is to that - probably me being more careful with superglue! I guess a little etching on the lid could create a gutter to stop the glue spreading into the slider, but that may be too much complexity?

Finally, the one drawback of my hook is that the holes in the lid are further out, so you can see through the part when it's taken off the box. I don't think that is a problem, but it's not quite as pretty as without it. Moving everything in by `t` would fix it, but I'm inclined not to unless you think it's important: keeping the latches close to the edge is good for strength.

rwb27:
Oh, and would it be appropriate to add a `--bottom` flag so I can have a latched plate both top and bottom? I would like to do this for the microscope box, though of course I'm subclassing anyway there. I suspect others might find a removable top and bottom useful, though.

florianfesti:
--bottom added.
I moved the latches back into the middle. They should still be more than strong enough. I want to use this code for other boxes, too. And some of them might not have as much brim.

rwb27:
Thanks :) Would it be helpful if I re-print a test box and check it? I will still have the issue of cutting order: shall I just pick a couple of other colours to use? I have not gone through many of the boxes, but I'm not aware of any standard colours you've used for that purpose elsewhere...

florianfesti:
Good you brought this up again. I forgot that I wanted to make the slider slightly smaller. That should solve the nesting issue (in theory). Ofc this will depend on the software you are using and whether it will do rounding in a range that is bigger than the current difference.

 Re-cutting would now be very helpful. You can get away with just cutting the lid pieces. The lid should still fit an older box of the same size.

rwb27:
> Good you brought this up again. I forgot that I wanted to make the slider slightly smaller. That should solve the nesting issue (in theory). Ofc this will depend on the software you are using and whether it will do rounding in a range that is bigger than the current difference.
> 
> Re-cutting would now be very helpful. You can get away with just cutting the lid pieces. The lid should still fit an older box of the same size.

Sorry for the delay in posting - I re-cut this last week and it looks good: the holes in the outer plate for the sliders now sit fully above the inner lid piece, so there's no unsightly gap, which is nice. The latches still work well, as before.

Maybe I didn't explain the nesting issue well enough: the problem is not that the cuts interfere with each other, it's that the cuts happen (on my machine) in the wrong order. If the latches get cut out before the holes in them, they shift or fall through the bars on my cutter, and so the holes either don't get cut, or get cut in the wrong place. Everywhere else in boxes.py, `@holeCol` makes sure the holes are cut first, followed by the outline - but this isn't sufficient for the latches, because we need to first cut the holes, then cut the latch sliders, then cut out the holes for the sliders, then cut the outline of the plate. I've mocked this up in Inkscape, with two possible solutions. I can post photos of the bits that went wrong if it helps?

Doesn't work:
image

Works but is a bit wasteful:
image

Would work but possibly uses too many layers?
image

rwb27:
I've pushed a commit that does the final option, in case seeing some code helps decide which is the better option. The output is below:

image

box

florianfesti:
OK, I finally did cut the latches myself. LightBurn at least is able to figure out the right cut order without your patch now that the lines are truly inside of each other.

But the assembly is so messy and risky. I think we need a different solution that doesn't require gluing moving parts in situ. I actually like the latches otherwise. Them not sticking out so they can't be bumped open easily is really nice.
But I guess we need an additional layer somewhere to get the glue away from the sliding surfaces.

I'll try to come up with something.

rwb27:
I've tried this again being slightly more careful with the glue, and it was fine - it could perhaps be improved by etching a line (or making a series of small holes) around the outside of the cut-out where the latches slide: that would hopefully soak up any excess glue and make the whole thing more forgiving. Obviously it doesn't solve the need to glue the sliders to the latches, but I didn't find that was a problem.

Is it the glue between the outer lid (with the holes for the sliders) and the inner lid (with the slots for the latches) that you reckon is problematic?

PR #833, opened 2025-10-20
Lids: Labels for "overthetop" and "ontop"

MartinScharrer:
This PR simply adds labels to the lid walls for the styles "overthetop" and "ontop" which are missing compared to the other styles.


florianfesti:
Thanks for the patch!

PR #835, opened 2025-10-22
Added `spacing` argument to allow to set the spacing between parts.

MartinScharrer:
This patch allows the user to modify the spacing between parts.

### Motivation
I am using LightBurn and use its kerf feature, so I usually set *burn* to 0 and export to lbrn2 format.
As I am using the "Remove overlapping lines" feature (with distance at least the size of the kerf) of LightBurn and usually place my parts one kerf from each other so that parallel edges are only cut once. With the default spacing of half a thickness I always need to adjust the placement manually.

This addition allows for the original default spacing but also to adjust it according to the thickness *and* also add an extra absolute amount.

I myself would set *burn* to 0 and *space* to `0 : 0.125` where 0.125 is my kerf. This places all part one kerf apart and LightBurn removes parallel edges, reducing the burn time and reduces material waste.


florianfesti:
Thanks for the patch!

PR #836, opened 2025-10-24
Update server URL print format to generate full URL

MartinScharrer:
This minor change ensures that the console shows the full, valid server URL which is clickable from some consoles, like in PyCharm.

Before: `BoxesServer serving on :8000/...`

After: `BoxesServer serving on http://localhost:8000/...`

This makes testing new generators locally a little simpler as it saves the developer to enter the URL manually every time.


florianfesti:
I think defaulting to localhost is not the same thing than not having a host name. AFAIK the later serves on all interfaces while localhost only serves, well, locally.

florianfesti:
Thanks!

PR #837, opened 2025-10-25
Sooo.... A lot of things. New inner corner dogbone/ new DXF generator / new generator toolbox/briefcase

vinicius795:
Before anything else, everything “I” did here was actually done by ChatGPT Codex in VS Code. I can program, but not that well, and I don’t have the time needed to learn and implement everything required to do what I’d like to do.

So… I don’t expect this PR to be accepted; I’d just like to get a code review. I want to split this PR into three different ones, but that will take me a while, and I think someone might find the changes I made useful and start using them before they officially make it into the project. I’ll explain each change.

**–‐inner_corner dogbone –‐D D –‐R R**

Although the project is meant to create boxes cut on laser cutters, the only thing that stops it from working on any CNC router is that the internal corners need to have a dogbone shape. So I did exactly that, because I don’t have a laser cutter but I do have a CNC router. Since the corner drawing process required a lot of code, I thought it was better to put everything in a separate file. It follows the same flow as any other --inner_corner value, but adds the --D and --R options so you can specify the diameter or radius of the dogbone arcs. If you choose --inner_corner dogbone, then you must provide either --D or --R.

DogBone Inner Corners

There was a bug where if the distance between one notch and the next was very small, less than 2×D, the corner drawings would overlap. But that’s already been fixed, and now they stop where the two drawings meet.

image

**DXF Generator**

Currently, to generate DXF files we use a PS‑to‑DXF converter. The problem is that it flattens all curves into line segments. At first that isn’t a problem, but if the computer isn’t powerful and the file is very large—especially with the dogbone option—opening the file to do anything can be difficult. Another thing is that G‑code generation software can benefit from the drawing having arc entities; instead of creating 20 G1 lines to draw each arc, it can create just one G2/G3. So I wrote a “native” generator using ezdxf, and I took the opportunity to implement some improvements:

- Cubic curves become Arc, Circle, or Spline entities
- Concentric arcs of the same radius are flattened into a single arc or a single circle
- Collinear lines are joined into a single line
- Closed drawings are joined into a single polyline

This way the DXF file becomes shorter and more efficient.
The only “bug” I couldn’t solve is when using --inner_corner loop: the loop is flattened into line segments, but if it isn’t flattened like that, it can’t be joined into the polyline. To do that, the polyline would need to be a spline, which would add a lot of complexity that I didn’t think was worthwhile.

I didn’t test the DXF generator with every generator in the project or with all edge types, so there may be cases where it produces a bug. But I tested all the scenarios I could think of.

**ToolBox / BriefCase Generator**

I’m undecided about the name for this generator because I made it to store my power tools, but when you look at it you think “this is a briefcase.”
It’s a box with hinges that opens at the middle of the value of --H, it has a handle and an internal divider. I still need to make a mechanism to lock it closed so it doesn’t open when you’re carrying it by the handle. I haven’t had time to build one yet, but in a few days I should be able to and will post a picture. Because I had to create some custom edges, there are some bugs with parameters that I still need to fix.

### DIFF
--- a/boxes/drawing.py
+++ b/boxes/drawing.py
@@ -1028,7 +1028,7 @@ class LBRN2Surface(Surface):
                             C, x, y = c[0:3]
                             if C == "M":
                                 if start and points_equal(start[1], start[2], x0, y0):
-                                    pl.text = "LineClosed"
+                                    pl.text += f"L{cnt-1} 0"
                                 start = c
                                 cnt = 1
                                 if self.dbg: print ("next, because M")
@@ -1056,6 +1056,8 @@ class LBRN2Surface(Surface):
                         if start and points_equal(start[1], start[2], x0, y0):
                                 if bspline == False:
                                     pl.text = "LineClosed"
+                                else:
+                                    pl.text += f"L{cnt-1} 0"
                         start = c
                         if self.dbg: print ("2", num)
                     elif C == "T":


florianfesti:
This indeed is a lot. I have to warn you that my track record of reviewing large patch sets is not great. Can you at least split this into separate branches and PRs?

If your git foo is not up to that I can also try.

vinicius795:
Sure i can, my skills in git are not grate, but i will try. i think i will start with the dxf generator. it introduces very litle changes to the main code. do you think i should keep it in a separeted file or merge it into the drawing file?

a lot of commits is just me using github as a "pendrive", i have my pensonal computer and my work computer, and i use the commits to keep the code synchronized between the PCs.

florianfesti:
Keep is separate for now. Now that we have 4 different backends I wonder if we should split everything into their own file instead of mashing them together

vinicius795:
I just created PR #843, It added the new DXF generator.

PR #838, opened 2025-10-25
New generator: Box to hold plastic cages for honey bee queens

MartinScharrer:
I created a new generator to produce a transport box which holds plastic cages for honey bee queens usable for bee keepers which do queen rearing. Such a box provides a save way to store and transport multiple bee queens. 

Air circulation is ensured by providing options for multiple air holes.

The cage type and numbes can be selected using parameters.

The generator also alows to cover the lid with a bee queen icon in order to ensure proper housing for the royal content :-)


florianfesti:
I wonder if this should have  a second floor to hold the cages more securely.

May be also duplicate the floors without finger joints with slightly smaller holes so they can be cut out from foam to hold the cages without rattling. Not sure how much that would put the limbs of the royalty in danger, though.

MartinScharrer:
> I wonder if this should have a second floor to hold the cages more securely.
A second floor is already supported and can be placed freely with `sh` paramter (just add third section and select layer2).

However, the cages are hold quite secure in my tests already. Similar cardboard boxes I saw only have one cutout layer as well.

If wanted one rectangular foam could be placed on top under the lit to push the cages from top. I do not think it needs cutouts.

florianfesti:
What's still missing here? Can we get this finished or is it waiting on something specific?

MartinScharrer:
From my point of view it could be published as it is.

PR #839, opened 2025-10-26
New generator: Bee queen cage

MartinScharrer:
This new generator produces a wooden cage to hold a honey bee queen cell plug and later the hatched queen.

### Description
Cage box to house a bee queen.
The default opening on top is suitable for a Nicot queen-rearing socket or a cork plug.
This makes the cage suitable as a hatching cage.

Holes can be configured per side. The default value of 3.0mm is suitable as air holes, but this can be changed to
produce queen excluders (4.2mm) or drone excluders (5.2mm). Some space should be left closed to provide cover for
the young queen against aggressive worker bees.


florianfesti:
I am honored that Their Majesties are considering traveling with Boxes.py.

I wonder if there are a bit too many settings here. May be one slit size per cage is enough and we can have a separate setting (probably a choice parameter) of which walls get holes (Not that I know much about bee keeping).

For the top layer of the plug you can use `Boxes.parts.wavyKnob` or `Boxes.parts.concaveKnob` for easier gripping.

I also wonder if there should be more secure way to fit the plug.

MartinScharrer:
Thanks for the feedback. Yes, the options where my concern as well. I will think about them again.

I also will look at the knobs.

florianfesti:
Ok, here a suggestion on only having one set settings for all holes.

MartinScharrer:
Thanks, I will check them out.

florianfesti:
OK; cleaned things up further. After squashing all the commits this should be pretty close to complete. I mean it's just a small little box, even if it is for Royalty.

florianfesti:
Squashed everything into one commit. IMHO this is ready to be pushed. May be have a final look before I press the button.

florianfesti:
OK, I'll just merge this now. Feel free to open a new PR if you want to adjust anything.

PR #840, opened 2025-10-27
Add line width argument for SVG preview and export

### LINKED ISSUE #607: line-width flag
Currently the SVG line width is set to `max(burn, 0.05)` in `boxes/__init__.py`. My laser engraver only cuts lines of 0.001mm, anything else is engraved. I'd love to be able to control burn-correction and line-width explicitly and independently :)

MartinScharrer:
This PR implements a line width argument which allows the setting of the SVG line with independent of the burn parameter.
This was requested originally by #607 as the user there seems to use a laser engraving software which needs a line width of 0.001mm for burning but engraves thicker lines instead. The current implementation enforces a minimum of 0.05mm.

I also like to set the burn parameter to 0 and use LigthBurn's kerf offset setting instead, but find the 0.05mm minimum hard to view on the SVG preview. I therefore would also like to see such a parameter.

I implemented it in a way, that the value is an empty string by default and then follows the original value of min(2x burn, 0.05) making the change fully backward compatible.

Fixes #607.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -408,6 +408,9 @@ class Boxes:
         defaultgroup.add_argument(
             "--output", action="store", type=str, default="box.svg",
             help="name of resulting file")
+        defaultgroup.add_argument(
+            "--line_width", action="store", type=(lambda x: float('NaN') if x == "" else float(x)), default="",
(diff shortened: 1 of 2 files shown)


MartinScharrer:
Fixes #607

PR #841, opened 2025-10-27
Added missing new lines reported by pre-commit hook

MartinScharrer:
The following four files miss the trailing newline. They are getting reported by the `pre-commit` script.

As I run the `pre-commit` hook on my other branches before creating pull requests this diff always pop up. It would be nice to have this minor fix upstream to avoid the noise.


florianfesti:
Damn it, I was too quick with the merge button.

This does not work. Those are symlinks under Linux and can't be changed like that. Someone needs to look at the differences between Linux, Windows and git in handling that and tell the pre-commit folks. But it's probably not going to be me.

florianfesti:
Removed the patch from master again.

MartinScharrer:
I added them under Windows, so I guess there where added as Windows newlines.
The newlines might need to be Linux newlines, I guess. I will check this again.

PR #842, opened 2025-10-28
Added symlinks to pre-commit excludes to avoid issues with missing trailing spaces under Windows

MartinScharrer:
Follow-up to PR _Added missing new lines reported by pre-commit hook_ #841.

This minor issue can be fixed to exclude the symlink files from the pre-commit check.


florianfesti:
Still kinda weird that pre-commit isn't able to figure out that those are symlinks. Thanks for the patch!

PR #844, opened 2025-10-31
Update CabinetHingeEdge

vinicius795:
I fixed the hinge part drawing builder where it was creating a backarc where it wasn't needed. I believe this is a much bigger problem, where when a polyline has a drawing of a line of size 0, an arc with radius > 0, and a line 0 followed by an arc of radius 0, the program will always generate this. It's a "problem" with the burn option. If it's off, this doesn't happen.

**What causes the problem**
--burn > 0
self.polyline(0,[180, 5],0,-90, ....)

image

**How to "Fix"**
self.corner(180,e)
self.moveTo(0,0,-90)
self.polyline( ..... )

image

**OR**
--burn 0

image

### DIFF
--- a/boxes/drawing.py
+++ b/boxes/drawing.py
@@ -887,7 +887,7 @@ class LBRN2Surface(Surface):
                             C, x, y = c[0:3]
                             if C == "M":
                                 if start and points_equal(start[1], start[2], x0, y0):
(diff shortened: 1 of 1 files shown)


vinicius795:
I closed it because it's part of a bigger problem.

PR #851, opened 2025-11-07
Added new drink carrier generator

ceakins:
This allows you to create a drink carrier for uber eats/dasher/ect traveling with drinks.


florianfesti:
Nice generator. A few remarks:

Using a `h` or `s` edge at the bottom will hold the bottom much stronger

The handle is not really attached strong enough. You need finger joints at least the the ends. Having one wall  on each side between the bottles may also help. Picture how the forces from the weight of the bottles flow through the construction to the handle. Right now there is just one finger joint that is pulled apart. The finger joints are much stronger  along the joint.

It looks like the sidewalls will collide with their fingers at the second layer. You should just have one wall at the sides and add finger holes in the middle. You'll need a callback for that.

florianfesti:
OK; I had the feeling this was still not sturdy enough. Now the handle is connected to the end walls. With the "h" edge on the bottom this should be able to carry full bottles.

Did you test cut this one already? I often cut a miniature first to not waste too much material if there still are issues. Like 5 or 7 cm in size - depending on the complexity of the model.

May be it now deserves an assembly instruction. I might add that when we agree this is the final design.
I also think the defaults can be tweaked. There is a lot of space between the bottles right now.

florianfesti:
In general I am not a fan of copying another design one by one. I still wonder if we can take a few inspirations from #854. Having the carrier more open allows to get rid of the corners and also makes it easier to see what kind of bottles are in there. It also allows for quicker drying if anything gets spilled.

PR #852, opened 2025-11-10
fix kamishibai issues wiht sliding doors

gpsqueeek:
fix kamishibai issues whit sliding doors

- front and back shapes not closed properly when using screws

- add margin to the top holes for the sliding doors)


florianfesti:
Only had a quick glance over the changes. I hope you know your code better than I...

Thanks for the fixes.

gpsqueeek:
I was actually trying to make quite a big change (adding an optional space below the pictures to allow electronics for LEDs for example) when I found those, so I did go back deeply into my code... After some thoughts I cancelled the big changes (might not be that useful in the end) but kept the fixes !

PR #863, opened 2025-12-10
Fix Matrix generator wall slots and correct param help text

sayhiben:
- Corrected Matrix walls to cut slot chains on the matching axis; it previously had long/short walls swapped
- Renamed and reordered PCB dimension parameters for clearer input and corrected help text
- Updated pot/zh_CN msgids to reflect the corrected parameter names

### DIFF
--- a/boxes/generators/matrix.py
+++ b/boxes/generators/matrix.py
@@ -56,8 +56,8 @@ class Matrix(Boxes):
 
     led_width: int
     led_height: int
-    pysical_led_y: int
-    pysical_led_x: int
+    physical_led_x: int
+    physical_led_y: int
 
     distance_between_leds: float
     plexiglass_thicknes: float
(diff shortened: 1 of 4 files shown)


boxes/generators/matrix.py
```diff
@@ -313,7 +313,7 @@ def render(self):
             label="Wall 3",
             callback=[
                 lambda: self.matrix_back_sideholes(
-                    self.pysical_led_y + 2 * self.matrix_front_frame_border_offset
+                    self.physical_led_x + 2 * self.matrix_front_frame_border_offset
```
sayhiben: See above comment about X != Y

po/zh_CN.po
```diff
@@ -7489,22 +7489,22 @@ msgstr "LED点阵中每一个灯珠的高度"
...
-msgid "pysical_led_x"
+msgid "physical_led_y"
 msgstr ""
 
-#. help for parameter pysical_led_x
+#. help for parameter physical_led_y
 #: boxes/generators/matrix.py
-msgid "Height of the LED matrix pcb in mm"
+msgid "Height (Y) of the LED matrix PCB in mm"
 msgstr "LED点阵pcb的高度(mm)"
```
sayhiben: I assume the existing translations are correct in zh-CN, but this corrects them in the source msgid so that the strings still match. Please update zh-CN as part of PR approval/deployment, as I do not speak this language.

florianfesti:
Thanks for the patch! I don't speak Chinese either. Those translations are 3rd party contributions. I'll just wait until someone picks up the change and fixes the message.

PR #868, opened 2025-12-21
Ventilation for electronics boxes

bnutzer:
Hi,

I am using the ElectronicsBox as an enclosure for a 100W power supply plus a micro controller to control an LED strip. Due to that, the box needs ventilation.

This PR adds options to configure ventilation openings to the lid.

What do you think? Is that worth supporting? My first box uses hexagonal holes, other hex and hbar/vbar versions are possible as well.

PS: PR contains updates to translations that are used elsewhere as well

09DAB646-92DB-4F41-AAF4-F046A6B763DB_1_105_c


florianfesti:
I am sorry, but this is not the way! There already is  bit of a mess with ventilation holes in the code that I plan to rip out. Then there is just too many ways to do the ventilation.
I think the right way is to have the users generate the patterns with the `HolePattern` generator. Any improvements to `HolePatterns` is very welcome.

bnutzer:
ACK. I'll close the PR.

Your recommendation is to create separate layouts, and merge them in the laser cutter software, do I understand that correctly?

Do you have a perspective on establishing a pattern for "universal callbacks" in the code? That would allow us to create derivatives of a generator by inheriting from it, and override just the required callback(s).

From a user's perspective: having to create separate layouts and merging them afterwards is extremely flexible, yet more time-consuming and error-prone. But I'm just a beginner :)

florianfesti:
I do have an universal callback thingie in a branch somewhere. The idea is that you can have thing like hole patterns independently of the (box) generator and just add it later on. But I haven't touched it in quite a while. I did not quite have the time and attention this year that the Boxes.py project deserves.

PR #878, opened 2025-12-29
New generator: Crate

markalex2209:
I created a new generator for creating crates.

It generates a stackable (optionally) box with handles. 
Additionally it can produce mask for cutout patterns in the walls. 

Image and thumbnail are added into PR.

I couldn't make pre-commit work on my setup at the moment, so it's possible I missed something in the code style. Sorry for that.


markalex2209:
Figured out pre-commit setup on my end =)
Fixed code formatting. Added missing example svg.

florianfesti:
Pretty neat generator. Added two small improvements.

It looks like you increased the pattern mask around the grip holes after taking the example image. Any particular reason for this? I kinda like the box in the image better.

markalex2209:
> Pretty neat generator. Added two small improvements.

Thank you.
I like corner finger extension. 
Regarding stabilizers - I'm not sure. But weather prevents me from creating a second box to actually test how stable is stacking without them, so I'll trust in your experience.

> It looks like you increased the pattern mask around the grip holes after taking the example image. Any particular reason for this? I kinda like the box in the image better.

Well, project for box in the image has been adjusted manually multiple times after pattern was added, with specific goal of making it "pretty". In the end handle there is lower then in a generated model.
I couldn't generalize which relations oh handle position correspond to that abstract "pretty" idea, even with two sets of sizes I have in mind.
And based on the relative nature of prettiness and the fact that if masks are generated, some manual edit of generated file will be needed, I placed masks in the highest possible position, and each user will be able to move them to their liking themselves.

florianfesti:
OK, I renamed the generator file to crate.py to match the name in the UI and squashed the commits.

florianfesti:
Thanks for your contribution!

markalex2209:
Thanks for cleaning up after me)

And great thanks for the project itself!

Issue #881, opened 2026-01-04
FlexBox3 got switched up dimensions


LorenzMende:
**Describe the bug**
The dimensions of the FlexBox3 are flipped/ mixed up.
This leads to confusion when creating the box, being the z "box height" parameter changing the width of the box and y "box depth" changing the height.

**To Reproduce**
Just use the default parameter of FlexBox3 and change either y or z and evaluate the result.

**Expected behavior**
The dimensions of the generated box should relate to the input parameters.

**Screenshots or Drawings**
(changing y to 200)
<img width="2344" height="1181" alt="Image" src="https://github.com/user-attachments/assets/911a55d9-fae1-4260-b25f-20b3b5df9ebf" />
(changing z to 200)
<img width="2399" height="1227" alt="Image" src="https://github.com/user-attachments/assets/cc25a008-3f64-4f35-aa3b-179f1b8ad3a9" />

**Additional context**
Is there some description of the default coordinate system?
Big kudos for this project, after writing some box-generators for myself I discovered your repo with much pleasure.

I already prepared the necessary changes and will create a PR.

florianfesti:
This is a part of #831. There are multiple generators that are modified copies of others and that did not get their dimension names adjusted to the new use case.

PR #882, opened 2026-01-04
corrected x,y,z of FlexBox3

LorenzMende:
I updated the base dimensions of the FlexBox3 to reflect the input parameters/ description.

Resolving #881


florianfesti:
Thanks a lot!

PR #888, opened 2026-01-20
Stevenson screen box generator

rwb27:
I run a project for our first years where they build a little weather station. Last year we made a 3D printed box, but this year we have a laser cutter. I have written a basic box generator that puts Stevenson screens on two sides of a box  (to allow airflow but keep out rain). I've tested it out (once so far) with 3mm plywood, and it worked nicely.

I realise this will need a bit of work before it's mergeable, but I wanted to share it as work in progress - if there's interest in improving and merging it, that would be super cool and I'll try to find time to improve it. If not, it was a fun exercise for me and I'll happily leave it there.

I attach a photo of the test box I made, and the SVG I cut it from.

A louvred weather station box.

box


florianfesti:
This is very neat. I would probably weather harden a few places more. Using a h edge for the floor would give you a nice drop edge that prevents water from clinging to the bottom after running down the sides.

Slanting the front to attach to the roof with an 90° angle will also prevent water from clinging to the bottom of the roof.

Not sure if using finger joints for attaching the roof is the right call. May be not offering any weak points to water is the play and just attaching the roof with glue. 

Fixed the placement of the trapezoidWall() upstream.

rwb27:
> This is very neat. I would probably weather harden a few places more. Using a h edge for the floor would give you a nice drop edge that prevents water from clinging to the bottom after running down the sides.

Thanks - that's nice to hear. It will be a while before I can get to this, but I'll try to make those improvements in due course. At the moment it's quite a theoretical exercise, but I agree finger joints aren't optimal for watertightness on the roof! Possibly the solution there is either sealing them with glue or similar, or adding an extra flat piece on the top (though that does probably invite water to wick between the two top pieces). I'm not sure which is the better option, or if you have other suggestions...

florianfesti:
Thinking about it a bit more this clearly is missing a way to access the instruments inside. I guess the best way would be to make the bottom screwed on. This basically turns the whole box into a hood on top of the floor plate with the instruments mounted on. The floor plate could also be narrower than the box to closer resemble the Stevenson screen which is also open on the underside.

florianfesti:
In case you haven't figured this out yourself: This is the kind of thing the debug checkbox was added for.

florianfesti:
OK; the box now looks roughly how I think it should. There are still a few things left to clean up. But they are more wrt code quality and should not change the resulting box.

May be you can have a look and cut one.

rwb27:
Wow, thanks! I was waiting for teaching to calm down before I got back to this - I'm out of town for a few days but will try to cut this when I'm back and give it a try.

rwb27:
I've cut the example SVG and it looks nice - I have only done a dry fit so far, and I think my kerf is a bit bigger than it should be so it's a loose fit, but otherwise it looks really good. Is there anything left for me to clean up? Should I stick the image below in the PR?

image

florianfesti:
Yes, please. There may be a couple of comments that are now outdated. Also pre-commit fails due the the way the StevensonScreenSettings are used. A page of the QueenBeeCage playbook (creating the actual Settings object in render()) should solve that.

rwb27:
I'm hoping I've tidied up the issues - pre-commit passes now and I hope I've added the images correctly!

florianfesti:
Looks like cf236473e57cbad89f9dc0557206836a84dd39ab changes (aka fixes) the placement of the alignment pin pieces.

For me this looks fine now. If there are no immediate plans I'd squash this into one commit and merge.

florianfesti:
Well, further changes can just go into a new commit. One thing I could think of is strengthening the floor with some ribs on the bottom that would also make mounting on a pole easier. But may be I am overthinging this.

rwb27:
Thanks very much for this :) I did have a nearly-finished commit to get the image in, but got stuck with the script that generated thumbnails - the script ran without errors, but didn't make a thumbnail. I see you have done that anyway, so I will not tinker further with ImageMagick versions.

The student project that motivated this looks like it won't run in the same form next year, which removes my main excuse for working on this - but if I do end up working on it some more I will of course send another PR. Thinking about pole mounting would be neat!

PR #892, opened 2026-02-17
Ramp generator for accessibility

clavierp-net:
Hello, this is a WIP for a generator ramp. It's advanced enough for me to start publishing. I will test in real size next week, and will do pictures then.
The dev was fun, integrated server is wonderfull. I had a lot of headache but the other generators provide a lot of solutions.


florianfesti:
Very nice work!

Rebased onto master to get the rectangularTriangles fix into master.

I might replace the Newton algorithm by a proper formula but I have not tried, so that might be more complicated than I am currently thinking.

I wonder if an option to also have finger joints on top of the inner triangles would make sense. We could even use some special finger joints that poke through to provide extra grip. Not sure if anyone really wants that much triangles for that to make sense, though.

clavierp-net:
Hi, sorry for the lack of response, my laptop broke and after various attempts to repair it, I gave up en bought a new one XD
For Newton, I could not find a solution after few hours of research, the equation seems unsolvable. A dichotomic search could ensure convergence, but I didn't find any parameters that failed the Newton.
For the inner triangles, it was more a structural fear as it's the part that takes most of the stress but after more tests, it seems very tough so it could make assembly more self constrained (as of now, I used nails to lock everything in place). Another solution would be putting only few joints just to manage the sliding of top part onto triangular parts.
I'll join few photos of the prototypes. I should do the installation soon as I might have finally everything.

clavierp-net:
IMG_20260228_161745

clavierp-net:
IMG_20260401_161700
Prototype and v1 side by side

clavierp-net:
IMG_20260401_161718
v1

clavierp-net:
I'll take better pictures for the thumbnail at the fablab (we have a booth with better lightning for better photos)

florianfesti:
Those look pretty nice.

Yeah, after trying to solve the formula by hand and passing it to Wolfram Alpha I have an increased appreciation for using the Newton method. My solution would probably just be to inscribe the ramp into the y/h triangle and use atan for getting the angle. That would allow extending the top layer to fill the triangle when grinding the ends to the right angle - similar to the FatBallDispenser.

Overall the generator looks pretty good already. If you add the images I am fine with just squashing the commits and merging the PR.

florianfesti:
Any updates on the pictures? I can also just add one of those above and get this merged.

clavierp-net:
Hello, I forgot to keep you updated. The lab is closed for vacation until next week so I think you can go with one of those previously posted.
The hard maths part was to ensure height without any grinding but after irl test, I'm not so rigid about it XD.
The real world is way less precise and I had to grind and add various bumps so the few mm added by the overhang is not very important.

florianfesti:
OK; here is the simplified calculation and one of the pictures added. If there are no more comments I am going to squash the whole branch into one commit and merge it the next few days.

PR #914, opened 2026-04-10
Devel brims

ranaur:
Added two new lid types: flatbrim ans fangbrim. Both are just like the flat type (the lid is a rectangle with the full size of the top os the box). The difference is that in flat lid, there is another rectangle slightly (thickness) smaller to glue, while in the flatbrim there are 4 a small finger walls thickness + brim_spacing smaller, so they can fit inside the box. In fangbrim there are only 4 "fangs" (small rectangles) on the corners of the lid to fit it indise the box.


florianfesti:
Great someone is working on the lids. They clearly deserve more love. 

Note that you don't need extra scripts to test them. Just use `scripts/boxes ABox --Lid_style=flatbrim` on the command line to render a box with lid.

I added the lids to the test suite (`pre-commit run -a`). I fully expect that to break with changes. Don't worry about that.

For the flatbrim lid I wonder if just two walls - one left and one right are actually enough. That would also allow them to be rounded over in at least one direction.

The fangbrim lid would probably benefit from a small recess that gives it a positive stop when sticking the pieces into the lid from below.

Both could grow a small spring that keeps the lid under tension. But that's not required to get this merged.

ranaur:
Hi Florian!

Should I write some test scripts? Where should I put them? 

Sure you need only two walls. The 4 walls are more for aestetics of the lid than functionality.

I can make another option to choose between styles (4 walls, 2 walls - vertical/horizontal) or even make a third option to make the "fang" brim as the third option (and removing the fang brim lid type). What do you prefer?

And lastly quoting Knuth's "Beware of bugs in the above code; I have only proved it correct, not tried it." :-) I executed it and measured in Inkscape, but I still didn't have time to rent the laser cutter to make the real boxes. So, some small adjusts may need to be made.

The idea of the sprint is good. Instead of making only the rectangular "fang" it can male a small "blunt arrow" shape (an L with a slanted point). It could even make a small hole (or even dent, controlling the intensity of the laser) in the wall to lock the lid. You'll need to push with a scrwedriver to unlock the lid. But I would need some test to work the sizes. Let me test the simple one first.

florianfesti:
You really don't need to write test scripts. [de5ad7a](https://github.com/florianfesti/boxes/pull/914/commits/de5ad7ac5cee323a9ad86c383af77c968ef2f1c2) adds test for all lid styles so far. The test fail on GH as c00f1e002c2e47d9e72038204872a8465c4c111b changes the way the file names are generated, but it passes locally. If you (pull), rebase and forcepush your branch I can adjust the file names.

For now I would avoid adding holes to the side walls. There isn't really infrastructure to do that nicely. I have a unfishished branch somewhere that allows passing an callback object to a box generator so that it has access to all parts. That would be great to have lids adding holes to the sides. So may be just hold up with that for now. Just sliding friction should already be a lot nicer than none.

ranaur:
Just did what you asked for.

ranaur:
DId a lot of improvements and cleanup after cutting the first boxes. Now it is working as expected. Should I open another PR or you can merge from this new point. Just let me know.

PR #919, opened 2026-04-21
Add 'upsidedown' option for part flipping along the X-axis

MartinScharrer:
The existing "mirror" value in the `where` argument in `Boxes.move()` allows to mirror the part, i.e. flip it left-to-right along the Y-axis. 

This pull request adds an additional value "upsidedown" which allows to flip the part upside-down, i.e. along the X-axis.

I personally need this for a generator which produces several identical parts where every other one should be flipped in order to optimize cut-outs. 

The needed code changes are minimal and no influence on existing code is expected.

### DIFF
--- a/boxes/__init__.py
+++ b/boxes/__init__.py
@@ -1240,14 +1240,15 @@ class Boxes:
         * "up" / "down"
         * "left" / "right"
         * "mirror"
+        * "upsidedown"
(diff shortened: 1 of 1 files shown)


florianfesti:
Yeah, this has been missing for some reason. Thanks for the patch!

PR #920, opened 2026-04-22
Made autorestart of boxesserver work under MS Windows

MartinScharrer:
With the current implementation the autorestart code uses the script file itself as executable.
This fails under Windows as the script is not a Windows executable.

This PR changes the restart command to use the python executable instead with the script name as an argument.

Tested under MS Windows. Please test under Linux before merge.


florianfesti:
Looks like it work!. Thanks for the patch!

PR #923, opened 2026-05-02
Fixes and improvements for LaserHoldfast

MartinScharrer:
Fix:
 * Corrected overcut at start - by adding an x move.

Improvements:
 * Added posibility to create multiple number of holdfasts
 * Added possibility to add a centered cut in the shaft, so it has some spring when pushed in


florianfesti:
Not too much of a fan of using the low level ctx methods here. For proper placement you can nowadays use polygoneWall() which draws a proper part and has support for callbacks.

MartinScharrer:
> Not too much of a fan of using the low level ctx methods here. For proper placement you can nowadays use polygoneWall() which draws a proper part and has support for callbacks.

Thanks for the hint. I updated the code now.

florianfesti:
Much better.

I wonder if having the cut as a percentage of `h` is a better UI. Not 100% sure though.

florianfesti:
OK, let's not overthink this. We can change that later if we really want to.

Thanks for the patches!

PR #928, opened 2026-05-16
fix the name of 6-sided platonic solid.

jabcross:
come on now


florianfesti:
Indeed. Thanks for the patch!

PR #930, opened 2026-05-17
Draft: Hobycase with single vertical wall

bofakin:
The Hobbycase is extended to allow to be build with only a single vertical wall.

To support all shelve positions combinations a TriFingerJoint edge is introduced. It has double length holes and the finger on its edges are organised to always fit the left or right side of an hole. This allow to insert such edges from both sides.

The changes to the Hobbycase are controlled by an new parameter which selects between the classic behaviour and the single wall mode. In  single wall mode, the double width holes are changed to single width and the number of vertical walls is reduced.

Current State: The generated files look good, but I have neither printed such a case nor assembled one in a 3D construction tool. The MR is mostly for early feedback to have everything ready for merge when the physical tests succeed.


bofakin:
I somehow messed up the first push - some changed were missing there. Now everything should be up there. Sorry for this.

PR #946, opened 2026-07-30
Frame loom

ojensen5115:
A while back I (manually) traced out vectors for the [Rocket Loom](https://www.popoutprojects.com/products/kitset-weaving-loom) which you can find [here](https://www.thingiverse.com/thing:7056512) 

However, I found myself wanting it to be parametrized, so I wrote a generator for it. 

This version is a bit "blockier" than the original because I couldn't for the life of me figure out how to use `self.curveTo` correctly. But the difference is only cosmetic, and after using this version for a bit I find myself preferring it anyway.

I parametrized it so that I could create a "bracelet-sized" loom, and currently this is the only version I have a photo for. I will include proper photos once I cut out and play with the default-sized version too. I have photos on Thingiverse, but the shapes are slightly different so it doesn't feel right to use them here.

Bracelet-loom photo:
PXL_20260730_104245333


ojensen5115:
whoops, after my last test cut I tried to fix the handle width to take material thickness into account, but after doing another test cut and finding they don't fit, I saw I got a sign backwards.

ojensen5115:
Here's some photos of the loom at default sizes. I'll update the PR to add the photos properly when I get the chance.

PXL_20260731_111903343
PXL_20260731_111249109
PXL_20260731_105223069

florianfesti:
This looks very nice and complete. From my POV this can be squashed and merged. Or is there anything you still want to tweak?

ojensen5115:
Thanks! I think it's good to go. There's a couple experimental ideas i still want to try, but i think we can add them as a follow-up PR if they end up working.

PR #947, opened 2026-07-30
Add a nix flake file

ojensen5115:
I'm not sure if you'd want this in your repo, but with it, any users of [NixOS](https://nixos.org/) or even just Nix the package manager (available cross-platforn) can simply run `nix develop` in the repo root to get a functional devshell with everything necessary installed. You can basically think of a dev shell as somewhat similar to a pyenv, but for all packages, not just python deps.

Personally I find it very handy.

image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,66 @@
+{
+  description = "Dev shell for Boxes.py (florianfesti/boxes)";
+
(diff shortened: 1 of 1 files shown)


florianfesti:
Sure, but be prepared that tickets about it will be assigned to you.

PR #948, opened 2026-07-31
UI enhancement: make whole button clickable

therobotreader:
This is an extremely minor UI change to take care of a pet peeve of mine - the whole button should be clickable, not just the words within.


florianfesti:
Hmm, guess the generator  entries in the Menu deserve a similar treatment. Otoh I wonder if we should just rip the menu interface out completely.

florianfesti:
No matter what: Thanks for the patch!

PR #951, opened 2026-08-10
Add Spanish (es) translation

erdanieee:
Adds Spanish (`es`) as a new interface language.

Two files, matching how `fr` and `de` were added:

| File | What it is |
| --- | --- |
| `po/es.po` | Source catalogue — 2418 of 2418 messages of the current template |
| `locale/es/LC_MESSAGES/boxes.py.mo` | Compiled catalogue |

**No code changes are needed.** Language discovery is entirely glob based —
`BServer.getLanguages()` globs `locale/*/LC_MESSAGES/boxes.py.mo`,
`setup.py` globs `po/*.po`, `MANIFEST.in` already covers `po/*`, and
`BServer.lang_re` already matches both `es` and `es-ES`, so
`Accept-Language: es-ES` resolves to this catalogue.

The compiled `.mo` is committed because `getLanguages()` looks for `.mo`
files, not `.po` files — without it the language never appears in the
selector. This is also why the committed catalogue matters in the Docker
image: `setup.py` shells out with `os.system(f"msgfmt {po} -o ...")`
without checking the return code, so on a machine without gettext the
build silently produces no catalogue at all.

### Translation decisions

* **Technical identifiers are left unchanged**: single-letter edge codes
  (`e`, `F`, `Š`, …), short section parameters (`x`, `y`, `h`, `sx`, `sh`, …),
  output formats (`svg`, `dxf`, `lbrn2`, …), pulley profiles (`GT2_2mm`, `MXL`, …)
  and pressure units (`Pa`, `mmHg`, …).
* **Only `<option>` labels are translated, never the values.** `arg2html`
  emits `<option value="{e}">{_(e)}</option>`, so form submission and the
  CLI keep working with the original English values.
* **Where a parameter name and its own help text describe different
  quantities, the label follows the help**, so the two agree on screen.
  Example: `--min_strength` is documented as *"strength of smallest wrench
  (in mm)"*, which is a thickness, so it is labelled *grosor mínimo* rather
  than a literal *resistencia mínima*.
* Neutral Spanish, impersonal/infinitive register for actions
  (*Generar*, *Descargar*, *Guardar en la URL*).

### Notes for review

* The catalogue is built against the committed template
  (`POT-Creation-Date: 2026-03-28`). Generators added after that date —
  FrameLoom — are not covered yet and will be picked up the next time
  `po/boxes.py.pot` is regenerated.
* The `.mo` was compiled with `pybabel compile` rather than `msgfmt`, so its
  header carries `Generated-By: Babel 2.10.3`. Regenerating it with `msgfmt`
  produces a functionally identical but byte-different file.
* The CLI does not translate argparse help for *any* language (`de`, `fr`,
  `zh_CN` behave the same), so nothing changes there. `get_translation()`
  does resolve correctly, so `LANGUAGE=es` reaches `box.translations`, which
  is what `UBox` uses for its engraved part labels.

### Verification

* `/?language=es` and `/TypeTray?language=es` render in Spanish; the language
  selector lists `es` **and marks it as selected** (the directory name and the
  catalogue's `Language:` header match, which is what that comparison relies on).
* Generating an SVG through the form still works, confirming that translating
  choice labels did not affect submitted values.
* The error page substitutes correctly: *"Error al generar Caja universal"*.
* `pytest` — 194 passed, 9 skipped, including the byte-for-byte
  `examples/*.svg` comparison; the suite also passes with `LANGUAGE=es` set,
  so no translated string leaks into rendered output.
* `pre-commit run --files po/es.po locale/es/LC_MESSAGES/boxes.py.mo` — clean.
* A consistency pass over the catalogue checks that every `%s`/`%d`, URL,
  image path, HTML tag and entity is preserved, and that no entry is marked
  fuzzy.


florianfesti:
Funny coincident that I am currently in Asturias for vacation. Thanks for the translation!

PR #953, opened 2026-08-20
New top options for basedbox

ranaur:
The original basedbox can only be closed. I added the option to choose changes in the top (default is the same behaviour). But now you can ommit the top (edge type e, make it clickable, and many more.


florianfesti:
This is a valid feature, but the details are already implemented in `lids.py`. Have a look at the [UniversalBox](https://github.com/florianfesti/boxes/blob/master/boxes/generators/universalbox.py) to see how to make use of that. You need to inherit `_TopEdge` and use [.topEdges()](https://github.com/florianfesti/boxes/blob/c41187d9e68b2c42662d0bf65536cbd14b4c7e9f/boxes/generators/universalbox.py#L63) and [.lid()](https://github.com/florianfesti/boxes/blob/c41187d9e68b2c42662d0bf65536cbd14b4c7e9f/boxes/generators/universalbox.py#L100).

PR #966, opened 2026-09-16
scripts: fix sys.path fallback to project root

gyscos:
The pathlib migration in #965 changed the import fallback in boxes2inkscape and boxes2pot from `os.path.dirname(__file__) + "../"` (the parent of the script's directory) to `Path(__file__).resolve().parent` (the parent of the script itself, which is the script's own directory).

The fallback import of boxes.generators then fails as well, so both scripts exit with ModuleNotFoundError.

Go up one more level, matching the previous behavior.

Thanks for this project!


florianfesti:
Thanks for the fix!

